PR #43, opened 2015-01-14
change automated test checkout branch from pulp-2.4 to pulp-2.5

beav:
This is a temporary fix to set the branch to 2.5. Ideally it would be read from
the build params when the job runs.


barnabycourt:
LGTM

PR #45, opened 2015-02-02
remove qpid persistent store

mhrivnak:
The removed package failed to work with this image when run on an Ubuntu
host. Since this is just a demo deployment of pulp, it's sufficient to
live without the persistent store.

I also cleaned up an ADD and WORKDIR statement that I don't think were
adding any value.


rbarlow:
LGTM.

PR #46, opened 2015-03-06
724 - Fix integration test runner

barnabycourt:
Fix the integration test runner & remove script to update images that is no longer required since we will reuse the images built by nodepool

closes #724

https://pulp.plan.io/issues/724


beav:
LGTM

PR #47, opened 2015-03-09
158 - Support multiple versions of the same package in external_deps

barnabycourt:


beav:
LGTM,  but see note about deep dive request in https://github.com/pulp/pulp/pull/1704.

PR #48, opened 2015-03-16
141 - Validate and keep up-to-date redmine and bugzilla data

bmbouter:
Adds a Jenkins job definition which periodically runs a python
script. The Python script validates that upstream and downstream
bugs have bi-directional relationships. The script also replicates
upstream issue status to the downstream external tracker info. A
report of issues identified and changes made are sent to a list
of admins.

### DIFF
--- a/ci/jobs/ci-redmine-bugzilla.yaml
+++ b/ci/jobs/ci-redmine-bugzilla.yaml
@@ -1,0 +1,62 @@
+# This Jenkins job runs a script which performs automation related to Redmine
+# and Bugzilla. See https://github.com/pulp/pulp_packaging/ci/redmine_bugzilla.py
+
+- job-template:
+    name: 'redmine-bugzilla-automation'
+    defaults: ci-workflow-runtest
+    node: 'f21-np'
+    scm:
+        - git:
+            url: 'https://github.com/pulp/pulp_packaging.git'
+            branches:
+                - origin/master
+            basedir: pulp_packaging
+            skip-tag: true
+            wipe-workspace: false
+    triggers:
+        - timed: "0,30 * * * *"
+    wrappers:
+        - credentials-binding:
+            - file:
+                credential-id: 099bc04b-80b6-4cd6-ad1d-11908641f539
+                variable: REDMINE_BUGZILLA_CONF
+        - timeout:
+            # Timeout in minutes
+            timeout: 30
+            timeout-var: 'BUILD_TIMEOUT'
+            fail: true
+    builders:
+        - shell: |
+            #!/bin/bash
+            git config --global user.email "pulp-infra@redhat.com"
+            git config --global user.name "pulpbot"
+            set -x
+            env
+
+            cd $WORKSPACE/
+
+            sudo pip install --upgrade git+https://github.com/bmbouter/python-bugzilla.git@expand-external-tracker-support
+            sudo pip install python-redmine
+
+            # Here due to InsecurePlatformWarning
+            # https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning
+            sudo pip install pyopenssl ndg-httpsclient pyasn1
+
+            python pulp_packaging/ci/redmine_bugzilla.py
+
+    publishers:
+      # Take the node offline so that another build doesn't pile on
+      - groovy-postbuild: "manager.build.getBuiltOn().toComputer().setTemporarilyOffline(true)"
+      - email-ext:
+          recipients: 'bbouters@redhat.com,mhrivnak@redhat.com'
+          reply-to: $DEFAULT_REPLYTO
+          subject: 'Upstream/Downstream Linking Issues'
+          body: $DEFAULT_CONTENT
+          failure: false
+          first-failure: true
+
+
+- project:
+    name: redmine-bugzilla-automation
+    jobs:
+     - 'redmine-bugzilla-automation'
--- a/ci/redmine_bugzilla.py
+++ b/ci/redmine_bugzilla.py
@@ -1,0 +1,172 @@
+"""
+This module does a few things:
+
+0) Verify that all Bugzilla bugs which point to a Redmine issues also have
+Redmine issues pointing back to Bugzilla bugs. This script print a report of
+changes made and any linking issues are found. If any link issues are
+detected, a RuntimeError is raised so the caller will know an issue occurred.
+
+1) For all Redmine issues that track Bugzilla bugs, ensure that the
+Bugzilla external tracker description, status, and priority all mirror the
+Redmine issue state. Also put a comment on the Bugzilla bug when changes are
+made.
+
+Requirements:
+The Redmine and Bugzilla credentials are expected from to be read from a
+file in ini format. See the example below. The location of this file is set
+by the REDMINE_BUGZILLA_CONF environment variable.
+
+<SNIP>
+[bugzilla]
+username = XXXX
+password = XXXX
+
+[redmine]
+key = XXXX
+</SNIP>
+"""
+
+
+import ConfigParser
+import os
+
+from bugzilla.rhbugzilla import RHBugzilla
+from redmine import Redmine
+import urllib3.contrib.pyopenssl
+
+# Here due to InsecurePlatformWarning
+# https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning
+urllib3.contrib.pyopenssl.inject_into_urllib3()
+
+
+BUGZILLA_URL = 'https://bugzilla.redhat.com'
+REDMINE_URL = 'https://pulp.plan.io'
+
+
+def get_bugzilla_connection(user, password):
+    """
+    Return the Bugzilla connection.
+
+    :param user: The username to connect with
+    :type user: basestring
+    :param password: The password to connect with
+    :type password: basestring
+
+    :return: An instantiated Bugzilla connection object.
+    """
+    return RHBugzilla(url='%s/xmlrpc.cgi' % BUGZILLA_URL, user=user, password=password)
+
+
+def get_redmine_connection(key):
+    """
+    Return the Redmine connection.
+
+    :param key: The api key to connect with
+    :type key: basestring
+
+    :return: An instantiated Redmine connection object.
+    """
+    return Redmine(REDMINE_URL, key=key)
+
+
+def main():
+    config = ConfigParser.ConfigParser()
+    config.read(os.environ['REDMINE_BUGZILLA_CONF'])
+
+    username = config.get('bugzilla', 'username')
+    password = config.get('bugzilla', 'password')
+    key = config.get('redmine', 'key')
+
+    redmine = get_redmine_connection(key)
+    BZ = get_bugzilla_connection(username, password)
+
+    redmine_issues = [issue for issue in redmine.issue.filter(query_id=24)]
+
+    non_closed_bug_with_ext_tracker = BUGZILLA_URL + '/buglist.cgi?bug_status=NEW&' \
+        'bug_status=ASSIGNED&bug_status=POST&bug_status=MODIFIED&bug_status=ON_DEV&' \
+        'bug_status=ON_QA&bug_status=VERIFIED&bug_status=RELEASE_PENDING&' \
+        'columnlist=priority%2Cbug_severity%2Cbug_status%2Cshort_desc%2Cchangeddate%2C' \
+        'component%2Ctarget_release%2Cassigned_to%2Creporter&f1=external_bugzilla.url&' \
+        'list_id=3309842&o1=substring&query_format=advanced&v1=pulp.plan.io'
+    bugzilla_bugs = BZ.query(RHBugzilla.url_to_query(non_closed_bug_with_ext_tracker))
+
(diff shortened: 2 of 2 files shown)


rbarlow:
I recommend amending your commit message, and including:

```
fixes #141
```

ci/jobs/ci-redmine-bugzilla.yaml
```diff
@@ -0,0 +1,62 @@
+# This Jenkins job runs a script which performs automation related to Redmine
+# and Bugzilla. See https://github.com/pulp/pulp_packaging/ci/redmine_bugzilla.py
+
+- job-template:
+    name: 'redmine-bugzilla-automation'
+    defaults: ci-workflow-runtest
+    node: 'f20-np'
```
rbarlow: Should we build this against f21 instead, since f20 only has a few months left?

ci/jobs/ci-redmine-bugzilla.yaml
```diff
@@ -0,0 +1,62 @@
...
+    scm:
+        - git:
+            url: 'https://github.com/bmbouter/pulp_packaging.git'
```
rbarlow: Shouldn't this use /pulp instead of /bmbouter?

ci/jobs/ci-redmine-bugzilla.yaml
```diff
@@ -0,0 +1,62 @@
...
+            branches:
+                - origin/141-add-bug-automation
```
rbarlow: And shouldn't it also use master?

ci/jobs/ci-redmine-bugzilla.yaml
```diff
@@ -0,0 +1,62 @@
...
+            basedir: pulp_packaging
+            skip-tag: true
+            wipe-workspace: false
+    triggers:
+        - timed: "0,30 * * * *"
+    wrappers:
+        - credentials-binding:
+            - file:
+                credential-id: 099bc04b-80b6-4cd6-ad1d-11908641f539
```
rbarlow: I assume this isn't a secret?

ci/jobs/ci-redmine-bugzilla.yaml
```diff
@@ -0,0 +1,62 @@
...
+        - shell: |
+            #!/bin/bash
+            git config --global user.email "pulp-infra@redhat.com"
+            git config --global user.name "pulpbot"
+            set -x
+            env
+
+            cd $WORKSPACE/
+
+            sudo pip install --upgrade git+https://github.com/bmbouter/python-bugzilla.git@expand-external-tracker-support
```
rbarlow: Shouldn't this also use /pulp instead of /bmbouter?

rbarlow: And also, master?

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
+"""
+This module does a few things:
+
+1) Verify that all Bugzilla bugs which point to a Redmine issues also have
```
rbarlow: What is this 1-based indexing nonsense?

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+[redmine]
+key = XXXX
+</SNIP>
+"""
+
+
+import ConfigParser
+import os
+
+from bugzilla.rhbugzilla import RHBugzilla
```
rbarlow: I didn't see the bash script in the yaml file installing this dependency. Should it be added there?

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+from redmine import Redmine
+import urllib3.contrib.pyopenssl
+
+# Here due to InsecurePlatformWarning
+# https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning
+urllib3.contrib.pyopenssl.inject_into_urllib3()
```
rbarlow: Nice, thanks for the link :bow:

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+BUGZILLA_URL = 'https://bugzilla.redhat.com'
+REDMINE_URL = 'https://pulp.plan.io'
+
+
+EMAIL_RECEIVERS = ['bmbouter@gmail.com']
+FROM_ADDRESS = 'bbouters@redhat.com'
+EMAIL_SUBJECT = 'Pulp upstream/downstream Bug Report'
```
rbarlow: This seems a little different than what was in the Yaml file. Does the YAML file override these?

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+    username = config.get('bugzilla', 'username')
+    password = config.get('bugzilla', 'password')
+    key = config.get('redmine', 'key')
+
+    redmine = get_redmine_connection(key)
+    BZ = get_bugzilla_connection(username, password)
+
+    redmine_issues = [issue for issue in redmine.issue.filter(query_id=24)]
+
+    non_closed_bug_with_ext_tracker = BUGZILLA_URL + '/buglist.cgi?bug_status=NEW&bug_status=ASSIGNED&bug_status=POST&bug_status=MODIFIED&bug_status=ON_DEV&bug_status=ON_QA&bug_status=VERIFIED&bug_status=RELEASE_PENDING&columnlist=priority%2Cbug_severity%2Cbug_status%2Cshort_desc%2Cchangeddate%2Ccomponent%2Ctarget_release%2Cassigned_to%2Creporter&f1=external_bugzilla.url&list_id=3309842&o1=substring&query_format=advanced&v1=pulp.plan.io'
```
rbarlow: This line is very wide, can we wrap for PEP-8 and readability?

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+    bugzilla_bugs = BZ.query(RHBugzilla.url_to_query(non_closed_bug_with_ext_tracker))
+
+    links_issues_record = ''
+    ext_bug_record = ''
+
+    for issue in redmine_issues:
+        for custom_field in issue.custom_fields.resources:
+            if custom_field['name'] == 'Bugzilla':
```
rbarlow: Is there a way to just access this without looping over all the fields?

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+                bug = BZ.getbug(int(custom_field['value']))
+                links_back = False
+                for external_bug in bug.external_bugs:
+                    if external_bug['type']['description'] == 'Pulp Redmine' and external_bug['ext_bz_bug_id'] == str(issue.id):
```
rbarlow: Wrap this line for PEP-8.

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+                        ext_params = {}
+                        if external_bug['ext_description'] != issue.subject:
+                            ext_params['ext_description'] = issue.subject
+                        if external_bug['ext_status'] != issue.status.name:
+                            ext_params['ext_status'] = issue.status.name
+                        if external_bug['ext_priority'] != issue.priority.name:
+                            ext_params['ext_priority'] = issue.priority.name
+                        if len(ext_params.keys()) > 0:
+                            ext_bug_record += 'Bugzilla bug %s updated from upstream bug %s with %s\n' % (bug.id, issue.id, ext_params)
```
rbarlow: This line is wide too.

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+                            ext_params['ids'] = external_bug['id']
+                            BZ.update_external_tracker(**ext_params)
+
+                        # Remove the bug list gotten via search so we don't examine it again when bugzilla_bugs are iterated through
```
rbarlow: Wide line.

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+                        bug_in_search_index_to_remove = -1
```
rbarlow: It would be cleaner to use None linguistically, instead of -1.

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+                        for i, bug_in_search in enumerate(bugzilla_bugs):
+                            if bug_in_search.id == bug.id:
+                                bug_in_search_index_to_remove = i
+                                break
+                        if bug_in_search_index_to_remove != -1:
```
rbarlow: If you used None here, it would read more naturally.

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+                            bugzilla_bugs.pop(bug_in_search_index_to_remove)
+
+                        links_back = True
+                if not links_back:
+                    links_issues_record += 'Redmine #%s -> Bugzilla %s, but Bugzilla %s does not link back\n' % (issue.id, bug.id, bug.id)
```
rbarlow: This line is too wide.

ci/redmine_bugzilla.py
```diff
@@ -0,0 +1,153 @@
...
+        for external_bug in bug.external_bugs:
+                if external_bug['type']['description'] == 'Pulp Redmine':
+                    issue_id = external_bug['ext_bz_bug_id']
+                    issue = redmine.issue.get(issue_id)
+                    links_back = False
+                    for custom_field in issue.custom_fields.resources:
+                        if custom_field['name'] == 'Bugzilla' and int(custom_field['value']) == bug.id:
+                            links_back = True
+                    if not links_back:
+                        links_issues_record += 'Bugzilla #%s -> Redmine %s, but Redmine %s does not link back\n' % (bug.id, issue.id, issue.id)
```
rbarlow: This line is too wide.

rbarlow:
LGTM.

ci/jobs/ci-redmine-bugzilla.yaml
bmbouter: Good idea; I switched it to f21

ci/jobs/ci-redmine-bugzilla.yaml
bmbouter: Good point. I need to change this just before I merge it.

ci/jobs/ci-redmine-bugzilla.yaml
bmbouter: Yes I need to change this just before I merge it. Good point.

ci/jobs/ci-redmine-bugzilla.yaml
bmbouter: Correct this is not secret.

ci/jobs/ci-redmine-bugzilla.yaml
bmbouter: It really should be `sudo pip install python-bugzilla` but we can't do that until my changes are accepted into a release that gets put on PyPi. I'm contributing them back to python-bugzilla, but until then it needs to stay like it is. I also forked from my private bmbouter account because it's not really a Pulp repo.

ci/redmine_bugzilla.py
bmbouter: It is in the bash script in the yaml file. It's the line:

`sudo pip install --upgrade git+https://github.com/bmbouter/python-bugzilla.git@expand-external-tracker-support`

ci/redmine_bugzilla.py
bmbouter: These are no longer used and have been removed. Good catch.

ci/redmine_bugzilla.py
bmbouter: I don't think so. It's an unordered list of dictionaries and I have to iterate to find the one whose index has the right data in its dict.

ci/redmine_bugzilla.py
bmbouter: Good suggestion. I originally used None, but when I checked the value I used `if bug_in_search_index_to_remove` which caused a bug where it wouldn't go in when `bug_in_search_index_to_remove` is 0. I opted for assigning it to `None` and evaluating using `foo is not None`.

ci/redmine_bugzilla.py
bmbouter: done!

bmbouter:
I so want to use the #fixes keyword, but I can't because it doesn't actually fully fix the issue. There are still a few non-code deliverables. I will be using that in the future. I updated the other changes I needed to make so I can merge.

rbarlow:
On 03/19/2015 04:35 PM, bmbouter wrote:
> I so want to use the #fixes keyword, but I can't because it doesn't
> actually fully fix the issue. There are still a few non-code
> deliverables. I will be using that in the future. I updated the other
> changes I needed to make so I can merge.

In that case, you can use this syntax:

re #141

It does all the same cool stuff as fixes (most importantly, tagging the
ticket with the specific commits that fix it), but it will not advance
the state or completion percentage of the ticket.

I think we should all tag our commits with at least re, when fixes and
closes don't make sense.

-- 
Randy Barlow

PR #56, opened 2015-05-13
Using the pulp-server-qpid group instead of just pulp-server

mhrivnak:
A change between 2.6.0 and 2.6.1 meant that python-gofer-qpid was no longer
getting installed while building the docker images. This Dockerfile should
have used the -qpid group to begin with, so it's not anyone else's fault.


mhrivnak:
I built the base image and its dependent images (worker and apache) with this change and using 2.6.1. Everything worked as expected in a quick smoke test.

PR #63, opened 2015-06-08
Add qpid-tools to base packages

abn:


mhrivnak:
Thanks! This workaround will not be necessary long-term, but it valuable for now. I'd hoped the fix in qpid would happen more quickly.

https://www.redhat.com/archives/pulp-list/2015-June/msg00000.html

PR #64, opened 2015-06-09
Update setup.sh

smapjb:
$DROOT is not known at runtime inside the container, this must be a copy and paste error.
Also required is the pull request from abn


mhrivnak:
Good catch. Thanks!

PR #74, opened 2015-07-23
Added src as an arch to run createrepo on

dkliban:


barnabycourt:
You want to make sure the comps file is not passed into the src repo

dkliban:
Already push this change to master. Including the change suggested by Barnaby.

PR #78, opened 2015-11-12
updating docker images to pulp 2.7

mhrivnak:


seandst:
Everything appears to have built with the right versions, and the end result is a working pulp/docker registry.

PR #79, opened 2015-11-24
Add ansible playbook for Pulp Smash automation

elyezer:
The pulp_smash_automation.yaml is a playbook that benefits of pulp and
pulp-smash roles. The former is a role which deploys pulp server and the
latter is a role which fetch latest Pulp Smash, configure it and run the
tests using py.test.

### DIFF
--- a/ci/ansible/ansible.cfg
+++ b/ci/ansible/ansible.cfg
@@ -1,0 +1,2 @@
+[defaults]
+remote_user=root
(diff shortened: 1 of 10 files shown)


elyezer:
Updated PR with some changes for RHEL. Got the following on a RHEL 7 machine:

```
========== 11 failed, 39 passed, 4 skipped, 22 error in 36.97 seconds ==========
```

I'm planning to check against RHEL 6 once I get a machine from beaker.

dkliban:
This looks good. I am going to merge. I will update the images we use for Jenkins so there are no conflicts with qpid packages.

elyezer:
@dkliban thank you very much.

PR #80, opened 2015-12-01
Add pulp_server and pulp_smash playbooks

elyezer:
Also remove pulp_smash_automation playbook to allow running Pulp setup
and Pulp Smash run separated.

In addition update Install packages for migration 009 task to be run on
RHEL6+ and to install python-qpid-qmf. Last but not least, remove
conditional from handlers since it is not needed.

### DIFF
--- a/ci/ansible/roles/pulp/tasks/pulp_server.yaml
+++ b/ci/ansible/roles/pulp/tasks/pulp_server.yaml
@@ -13,12 +13,10 @@
 - name: Install qpid-cpp server
   action: "{{ ansible_pkg_mgr }} name={{ item }} state=latest"
   with_items:
+    - python-qpid-qmf
     - qpid-cpp-server
     - qpid-cpp-server-linearstore
-
-- name: Install package for migration 0009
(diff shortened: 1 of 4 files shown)


ci/ansible/roles/pulp/tasks/pulp_server.yaml
```diff
@@ -16,9 +16,12 @@
...
 
-- name: Install package for migration 0009
-  action: "{{ ansible_pkg_mgr }} name=qpid-tools state=latest"
-  when: ansible_distribution == "RedHat" and ansible_distribution_major_version|int == 7
+- name: Install packages for migration 0009
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=latest"
+  when: ansible_distribution == "RedHat" and ansible_distribution_major_version|int >= 6
+  with_items:
+    - python-qpid-qmf
+    - qpid-tools
```
dkliban: How come you only want to install these packages for RHEL 6 and 7? What installs them for Fedora?

PR #83, opened 2015-12-07
Add job definitions for running Pulp Smash

elyezer:
Add job definition to run Pulp Smash against Fedora 22 and 23 and RHEL 6
and 7.

In order to have a very similar environment to a real use of Pulp the
jobs are designed to install Pulp on one machine and run Pulp Smash from
another machine. Translating to Jenkins, Pulp is installed on one slave
and Pulp Smash is run from other slave, at the end the jUnit report will
be published.

### DIFF
--- a/ci/jobs/pulp-dev.yaml
+++ b/ci/jobs/pulp-dev.yaml
@@ -1,0 +1,90 @@
+- scm:
+    name: pulp-packaging
+    scm:
+        - git:
+            url: https://github.com/pulp/pulp_packaging.git
+            branches:
+                - '*/master'
+            skip-tag: true
+
+
+- job-template:
+    name: 'pulp-{pulp_version}-dev-{os}'
+    concurrent: true
+    node: '{os}-np'
+    scm:
+        - pulp-packaging
+    wrappers:
+        - config-file-provider:
+            files:
+                - file-id: rhn_credentials
+                  variable: RHN_CREDENTIALS
+    builders:
+        - shell: |
+            sudo yum install -y git ansible libselinux-python
+            echo 'localhost' > hosts
+            source ${{RHN_CREDENTIALS}}
+            ansible-playbook -i hosts ci/ansible/pulp_server.yaml \
+                -e pulp_version=2.7 \
+                -e "rhn_username=${{RHN_USERNAME}}" \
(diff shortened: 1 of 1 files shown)


ci/jobs/pulp-dev.yaml
```diff
@@ -0,0 +1,90 @@
...
+    scm:
+        - git:
+            url: https://github.com/pulp/pulp_packaging.git
+            branches:
+                - '*/master'
+            skip-tag: true
+
+
+- job-template:
+    name: 'pulp-{pulp_version}-dev-{os}'
```
elyezer: Please let me know if you agree with this name.

ci/jobs/pulp-dev.yaml
```diff
@@ -0,0 +1,90 @@
...
+            which-build: specific-build
+            build-number: ${{TRIGGERED_BUILD_NUMBER_PULP_SMASH_RUNNER}}
+            flatten: true
+    publishers:
+        - junit:
+            results: report.xml
+
+
+- job:
+    name: pulp-smash-runner
```
elyezer: Please let me know if you agree with this name.

ci/jobs/pulp-dev.yaml
```diff
@@ -0,0 +1,90 @@
...
+            ansible-playbook -i hosts ci/ansible/pulp_smash.yaml \
+                -e pulp_smash_baseurl=${BASE_URL} \
+                --connection=local
+    publishers:
+        - archive:
+            artifacts: 'ci/ansible/pulp-smash/report.xml'
+
+
+- project:
+    name: pulp-dev
```
elyezer: Please let me know if you agree with this name. I will update yaml name to match this one.

The project name will not be presented on jenkins, just the job and job-templates names.

elyezer:
@dkliban do you have any feedback on this? I'm looking for these jobs be run after the packaging generation ones.

rbarlow:
This looks fine to me, but I think @dkliban should agree before merging.

dkliban:
It looks good to me also. LGTM.

dkliban:
Though we will need to change the pulp-smash job after #84 is merged. We'll want to make it a parameterized job that will have a Version parameter. It will then use that parameter to set a variable for ansible.

PR #84, opened 2015-12-09
Add pulp_smash_version variable

elyezer:
The variable is used in order to tell Pulp Smash which Pulp Server
version it is running against.

### DIFF
--- a/ci/ansible/roles/pulp-smash/templates/settings.j2
+++ b/ci/ansible/roles/pulp-smash/templates/settings.j2
@@ -2,6 +2,9 @@
     "default": {
         "base_url": "{{ pulp_smash_baseurl }}",
         "auth": ["{{ pulp_smash_username }}", "{{ pulp_smash_password }}"],
+        {% if pulp_smash_version %}
+        "version": "{{ pulp_smash_version }}",
+        {% endif %}
         "verify": {{ pulp_smash_verify }}
     }
 }


Ichimonji10:
It's a good idea to provide the "version" attribute, but it's not required. From `python -m pulp_smash`:

> By default, Pulp Smash assumes that the Pulp server under test is the absolute latest development version, but the `version` option can be used to explicitly run tests suitable for an older version of Pulp.

Here's how that's implemented:

```python
    def __init__(self, base_url=None, auth=None, verify=None, version=None):  # noqa
        …
        if version is None:
            self.version = Version('1!0')
        else:
            self.version = Version(version)
        …
```

elyezer:
@Ichimonji10 I have updated the PR to match the optional behavior, will test the playbook to make sure it is working.

Ichimonji10:
:+1:

ci/ansible/roles/pulp-smash/defaults/main.yml
```diff
@@ -9,3 +9,9 @@ pulp_smash_password: "admin"
 
 # If SSL certificate should be verified
 pulp_smash_verify: "false"
+
+# Pulp Server version
+# pulp_smash_version variable is optional, if you want to specify it then you
+# can do so by passing additional variables to ansible-playbook command or by
+# uncommenting the following line and editing as needed.
+# pulp_smash_version: 2.7
```
seandst: I think it might be better to leave this uncommented, and don't recommend changing this value, since every other mechanism for defining vars overrides this one.

I'm not entirely clear on what this value is. How is it different from [pulp_version](https://github.com/pulp/pulp_packaging/blob/master/ci/ansible/roles/pulp/defaults/main.yaml#L10)?

elyezer: @seandst I don't have an opinion regarding leaving it documented or not. So whatever you think is better works for me.

The difference is that this is for pulp smash and not for pulp itself. With this value you tell Pulp Smash what Pulp server version you are testing against. Pulp Smash is the automation framework used to test Pulp.

seandst: Well, the defaults are there so that you've got a default value, so commenting it out is confusing to me.

Wouldn't the version of pulp you're testing against be something you can get from a custom fact?

elyezer: @seandst grabbing that from a config file will even help testing if the Pulp Server is reporting the right version. This was a design decision within Pulp Smash I'm just replicating it to the play book that runs it.

Also regarding the default with comment this is something optional and Pulp Smash will consider latest Pulp Server if not version is provided. I left the comment since this is the first place I would check for expected variables on a playbook role. Worth mention that I'm learning Ansible and it could be a better place to document it.

Ichimonji10: > @seandst grabbing that from a config file will even help testing if the Pulp Server is reporting the right version. This was a design decision within Pulp Smash I'm just replicating it to the play book that runs it.

As the system under test, I don't trust Pulp. I don't even trust it to correctly state its own version, nor do I trust it to be packaged correctly. There may also be times when we want to run the test suite against a different version of Pulp - say, to see how much difference there is in compatibility between two versions of Pulp.

seandst: So make it an interactive fact, or have a sane default all the time. :)

seandst: The main issue is that the defaults file is the place where you put defaults. Defaults should not be optional, they should be easy to override, and they should be in revision control. I think that encouraging users to alter the defaults file is bad practice.

dkliban: So how should the version be specified for these jobs that are running pulp-smash?

Ichimonji10: If no value is given, then Pulp Smash will assume that Pulp is the absolute latest development version. Do you think that's a sane default? We could do something with an interactive fact as you mention, though I don't know if either elyezer or I know how to do that.

seandst: > So how should the version be specified for these jobs that are running pulp-smash?

I think that depends a lot on how ansible is being called. For example, we override a default value in our [Vagrantfile](https://github.com/pulp/pulp/blob/master/Vagrantfile.example#L40), because that's the best place for customizing ansible's behavior when vagrant calls it.

I've never worked with ansible in a CI setting so I don't know enough to say what's likely to be easiest.

> If no value is given, then Pulp Smash will assume that Pulp is the absolute latest development version. Do you think that's a sane default? We could do something with an interactive fact as you mention, though I don't know if either elyezer or I know how to do that.

It seems like the default should be a null (or nil) value, rather than not setting the var. And yes, I think that if it's not specified then defaulting to the latest is a great choice.

seandst: The interactive prompt is probably out of the question since I assume we want to automate this, but here a link for reference:

https://docs.ansible.com/ansible/playbooks_prompts.html

dkliban: What if we don't have a default at all for this variable?

Ichimonji10: What happens depends on this code: 

    {% if pulp_smash_version %}
    "version": "{{ pulp_smash_version }}",
    {% endif %}

If I remember my time with jinja correctly, references to non-existent variables are interpreted as being `None`. If that's the case, then no `"version": "…",` is placed in the config file. And that's fine.

> The `verify` and `version` keys are completely optional. … By default, Pulp Smash assumes that the Pulp server under test is theabsolute latest development version, but the `version` option can be used to explicitly run tests suitable for an older version of Pulp.

elyezer: Yes, not providing a value the if block will be skipped.

About a sane default in this case is without any value, but I still feeling that we should provide documentation about the version somewhere. The point now is, where to have that doc?

The playbook is being called on the job using `ansible-playbook -e pulp_smash_version=2.7 -e other_override_or_variable=value ...`. It will don't matter if the version variable is commented or have a sane default it will have the value specified by the value.

PR #85, opened 2015-12-11
Pass Pulp version to pulp-smash-runner job

elyezer:
Pass the Pulp version to pulp-smash-runner job as both Pulp Smash and Ansible Pulp Smash Playbook accept it.


dkliban:
looks good to me

PR #89, opened 2015-12-22
Fix the ok test trigger to be more appropriate.

rbarlow:


ci/jobs/ci-continuous.yaml
```diff
@@ -43,7 +43,7 @@
             - pulp
           # Poll once per minute since our jenkins isn't public and web hooks aren't an option
           cron: '* * * * *'
-          trigger-phrase: 'ok test'
+          trigger-phrase: 'ok jenkies'
```
dkliban: I don't think you spelled it right.

seandst: agreed...test is still spelled "test"

rbarlow:
lol

PR #93, opened 2016-02-10
add IRC notification to select jobs

seandst:
notifications added to platform PR test, plugin PR test, and build
automation jobs. I also threw in on the pulp-smash jobs, because why
not? :)


dkliban:
Pulp smash jobs usually get run as part of the pulp-2.8-dev-[f22, f23, rhel6, rhel7] jobs. The pulp-smash results are actually published with the first jobs. So we don't need to put notifiers on the pulp-smash-runner job.

PR #102, opened 2016-02-29
Converts pulp-[version]-dev-[platform] jobs to run ansibl over ssh

dkliban:


seandst:
maybe slap a comment on there explaining why that's needed, up to you

PR #104, opened 2016-03-03
Add pulp-coverage role

elyezer:
This role is responsible for installing and uninstalling the coverage
report capturing on a Pulp installation. It also generates and captures
the coverage report.


seandst:
Woooooooooooooooooooot!

PR #106, opened 2016-03-03
Capture coverage information for Pulp 2.8

elyezer:
Install the pulp coverage when installing Pulp 2.8 and include the
cobertura publisher in order to publish coverage results for Pulp 2.8
jobs.

### DIFF
--- a/ci/jobs/pulp-dev.yaml
+++ b/ci/jobs/pulp-dev.yaml
@@ -36,7 +36,10 @@
                 -e "rhn_username=${{RHN_USERNAME}}" \
                 -e "rhn_password=${{RHN_PASSWORD}}" \
                 -e "rhn_poolid=${{RHN_POOLID}}" \
-                -e "rhn_atomic_poolid=${{RHN_ATOMIC_POOLID}}" \
+                -e "rhn_atomic_poolid=${{RHN_ATOMIC_POOLID}}"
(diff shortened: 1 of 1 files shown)


ci/jobs/pulp-dev.yaml
```diff
@@ -36,7 +36,10 @@
                 -e "rhn_username=${{RHN_USERNAME}}" \
                 -e "rhn_password=${{RHN_PASSWORD}}" \
                 -e "rhn_poolid=${{RHN_POOLID}}" \
-                -e "rhn_atomic_poolid=${{RHN_ATOMIC_POOLID}}" \
+                -e "rhn_atomic_poolid=${{RHN_ATOMIC_POOLID}}"
```
elyezer: This is a fix from a previous change.

ci/jobs/pulp-dev.yaml
```diff
@@ -36,7 +36,10 @@
...
+            if [ "{pulp_version}" = "2.8" ]; then
```
elyezer: This will result in `if [ "2.8" = "2.8" ]; then` for pulp 2.8 and `if [ "2.7" = "2.8" ]; then` for pulp 2.7, it will look weird, but since we are generating all the jobs from the same source, there is no better way to handle this, at least AFAIK.

PR #108, opened 2016-03-10
Install coverage using pip

elyezer:
Use pip to install coverage Python package in order to ensure the latest
version. This is needed because on RHEL the version available is 3.x
while the version needed for the coverage hook is 4+.


elyezer:
@dkliban can you please sanity check this?

PR #111, opened 2016-03-10
leading zero on permission to prevent unexpected results

seandst:


elyezer:
LGTM

seandst:
The commit message is referencing the wording of the "mode" argument to the file module in this doc:
https://docs.ansible.com/ansible/file_module.html

It says: "For those used to /usr/bin/chmod remember that modes are actually octal numbers (like 0644). Leaving off the leading zero will likely have unexpected results."

We left off the leading zero, and were indeed seeing unexpected results. :)

PR #112, opened 2016-03-18
Change back pulp-smash-runner to run parallel

elyezer:
Get results faster by running more than one pulp-smash-runner.


ci/jobs/pulp-dev.yaml
```diff
@@ -10,7 +10,6 @@
 
 - job-template:
     name: 'pulp-{pulp_version}-dev-{os}'
-    concurrent: false
```
elyezer: Default is false, better to remove this line.

rbarlow:
LGTM, but I'll wait for a second ack from @dkliban or @seandst.

PR #113, opened 2016-03-23
Add more entries to subjectAltName on pulp-certs

elyezer:
This should fix the failures related to Docker cli on Pulp Automation Jobs.

### DIFF
--- a/ci/ansible/roles/pulp-certs/tasks/main.yml
+++ b/ci/ansible/roles/pulp-certs/tasks/main.yml
@@ -20,13 +20,13 @@
 - name: Configure openssl subjectAltName
   lineinfile: dest=/etc/pki/tls/openssl.cnf
               insertafter="^\[ usr_cert \]"
-              line="subjectAltName=DNS:{{ansible_fqdn}}"
+              line="subjectAltName=DNS:{{ ansible_fqdn }},DNS:{{ ansible_nodename }},DNS:{{ ansible_hostname }}"
 
 - name: Create Apache certificate request
   command: >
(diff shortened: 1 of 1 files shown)


ci/ansible/roles/pulp-certs/tasks/main.yml
```diff
@@ -20,13 +20,13 @@
 - name: Configure openssl subjectAltName
   lineinfile: dest=/etc/pki/tls/openssl.cnf
               insertafter="^\[ usr_cert \]"
-              line="subjectAltName=DNS:{{ansible_fqdn}}"
+              line="subjectAltName=DNS:{{ ansible_fqdn }},DNS:{{ ansible_nodename }},DNS:{{ ansible_hostname }}"
```
jeremycline: This looks reasonable to me.

Out of curiosity, do you know what the difference might be between the nodename and the fqdn? In my brief search of the Ansible documentation I wasn't able to find out the exact difference.

elyezer: I am not sure the real difference and the source of that information but on a system I was trying I got:

```
        "ansible_nodename": "rhel7-vanilla-np-qeos-88822.slave.openstack.org.novalocal",
        "ansible_fqdn": "rhel7-vanilla-np-qeos-88822.novalocal",
```

Seems that `ansible_fqdn` is `ansible_hostname.ansible_domain` and `ansible_nodename` is what `hostname` command reports.

jeremycline: Oh, weird

elyezer:
Let's merge and see what happens.

PR #115, opened 2016-03-24
Add subscription cleanup on RHEL machines

elyezer:
Make sure to release the subscription to avoid run out of subscriptions.

### DIFF
--- a/ci/jobs/pulp-dev.yaml
+++ b/ci/jobs/pulp-dev.yaml
@@ -18,6 +18,10 @@
             files:
                 - file-id: rhn_credentials
                   variable: RHN_CREDENTIALS
+        - inject:
+            properties-content: |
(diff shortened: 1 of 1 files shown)


ci/jobs/pulp-dev.yaml
```diff
@@ -69,13 +73,17 @@
             build-number: ${{TRIGGERED_BUILD_NUMBER_PULP_SMASH_RUNNER}}
             flatten: true
         - shell: |
-            if [ "{pulp_version}" = "2.8" ]; then
+            if [ "${{PULP_VERSION}}" = "2.8" ]; then
```
elyezer: Converted this a environment variable to avoid seen on jobs something like `if [ "2.8" = "2.8" ]` hard coded.

PR #116, opened 2016-03-29
reorganize jjb definitions

seandst:
This brings our JJB definitions in-line with the layout recommended by
JJB itself. Great care was taken to make no changes in the actual output
of JJB so that this PR can be reasonably tested.

To test:
- check out the parent of this commit
- run `jenkins-jobs test` against jobs defined in this directory, write
  the output xml to a file
- update back to this commit
- run `jenkins-jobs test` against jobs defined in this directory, write
  the output xml to a different file than before
- diff the two files, see that they are the same (If they aren't, let me know)


ci/jobs/README.md
```diff
@@ -0,0 +1,29 @@
+When defining jobs in this file, please organize files according to JJB's recommended layout:
```
elyezer: s/this file/this directory?

ci/jobs/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+http://docs.openstack.org/infra/system-config/jjb.html#configuring-projects
+
+The summary of what this means:
+- Job defaults go in `defaults.yaml`
+- Macros go in `macros.yaml`
+- Templates go in in files named similarly to the jobs they create so we can easily find them
+- Jobs using those templates are defined in `projects.yaml`, and the name of the project should
+  match the name of the file containing that project's jobs.
```
elyezer: Worth mentioning to use `-` as separator for yaml files? This will ensure a pattern on the name. On robottelo-ci I see files using both `-` and `_` as separator and sometimes in the same file name.

ci/jobs/projects.yaml
```diff
@@ -0,0 +1,110 @@
...
+     - 'build-automation-repo-{release_config}'
+    release_config:
+     - 2.6-dev
+     - 2.6-testing
+     - 2.6-release
+     - 2.7-dev
+     - 2.8-dev
+     - master
+
+- project:
```
elyezer: Not sure if I like this way of defining jobs, since we are "duplicating" code. Creating a project for an actual job definition seems reinventing the wheel because there is no practical difference of having a job definition. On the other hand projects are required for job-template definitions and in this case required to exist.

elyezer:
ACK, pending some comments.

ci/jobs/README.md
seandst: It think it might be, at least as a recommendation.

ci/jobs/README.md
seandst: Yup.

ci/jobs/projects.yaml
seandst: I agree, but think it's worthwhile to do it this way in the scope of this PR. The goal with the projects file in this PR was simply "Make sure the projects file points at every job file", and not "Reorganize the job files". That comes next. :)

seandst:
Thanks for the comments, I believe they have been addressed.

elyezer:
LGTM

seandst:
If anyone has to deal with conflicts as a result of what I've done here, I apologize for the inconvenience. Let me know if you want help resolving them.

PR #118, opened 2016-03-31
Capture logs for every Pulp Smash job run

elyezer:
The build status becomes unstable or failed when the jUnit report is publish.
Because that, the logging were not being captured in any circumstance.

### DIFF
--- a/ci/jobs/pulp-dev.yaml
+++ b/ci/jobs/pulp-dev.yaml
@@ -81,22 +81,17 @@
             steps:
                 - sonar:
                     sonar-name: 'Sonar Test Server'
-        - conditional-step:
-            condition-kind: current-status
-            condition-worst: FAILURE
-            condition-best: UNSTABLE
-            steps:
-                - shell: |
-                    rm -rf logs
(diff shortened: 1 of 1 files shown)


ci/jobs/pulp-dev.yaml
```diff
@@ -81,22 +81,17 @@
             steps:
                 - sonar:
                     sonar-name: 'Sonar Test Server'
-        - conditional-step:
-            condition-kind: current-status
-            condition-worst: FAILURE
-            condition-best: UNSTABLE
```
Ichimonji10: Aha. So formerly, we'd only capture logs if the build result was one of UNSTABLE or FAILURE. But that leaves out the other states of SUCCESS, NOT_BUILD and ABORTED, and there's value in capturing logs in those other states too. Right?

Reference: http://docs.openstack.org/infra/jenkins-job-builder/builders.html#builders.conditional-step

Ichimonji10:
ACK. I see value in capturing logs for all test runs, not just those that end in UNSTABLE or FAILURE.

PR #119, opened 2016-04-01
Update nodepool configuration and scripts

elyezer:
Make sure that the scripts and configuration matches the current state.

### DIFF
--- a/ci/nodepool/scripts/prepare_node.sh
+++ b/ci/nodepool/scripts/prepare_node.sh
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+HOSTNAME="$1"
+VANILLA="$(grep vanilla 2>/dev/null <<< ${HOSTNAME})"
+DOCKER="$(grep docker 2>/dev/null <<< ${HOSTNAME})"
 
 echo "Disable selinux"
 sudo setenforce 0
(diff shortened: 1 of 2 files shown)


ci/nodepool/scripts/prepare_node.sh
```diff
@@ -49,22 +51,35 @@ git clone https://github.com/pulp/pulp_packaging.git
...
-if  [ "$OS_NAME" == "RedHatEnterpriseServer" ] && [ "$OS_VERSION" == "6" ]; then
-    echo "cleaning up rhel6"
+if [ "$OS_NAME" == "RedHatEnterpriseServer" ] && [ "$OS_VERSION" == "6" ] && [ ! "${DOCKER}" ]; then
     sudo rm -f /etc/udev/rules.d/70*
     sudo sed -i '/^\(HWADDR)\|UUID\)=/d' /etc/sysconfig/network-scripts/ifcfg-eth0
 fi
+
+if [ "${DOCKER}" ]; then
+    sudo yum install -y docker
+    sudo systemctl enable docker
```
Ichimonji10: This will start docker on the next boot, but not right away. This is what you want?

elyezer: Yes, this script is used on image generation, that said docker running is not necessary at this time because the actual slave is going to be created by cloning the base image and then booting it.

So, when the actual slave is started the docker service will be enabled and then will start running.

Ichimonji10: :+1:

PR #121, opened 2016-04-05
Pull --skip-broken when installing packages

elyezer:
Unit test jobs should not `--skip-broken` when installing packages.


bmbouter:
thanks @elyezer

PR #122, opened 2016-04-05
Skip coverage report on RHEL6

elyezer:
The Python version available on RHEL6 is Python 2.6 which is a really old
version and the coverage report doesn't run on that version. Is better to skip
capturing the coverage report on that OS version in order to avoid the build to
fail and not publish the jUnit XML report.

### DIFF
--- a/ci/jobs/pulp-dev.yaml
+++ b/ci/jobs/pulp-dev.yaml
@@ -63,7 +63,10 @@
             build-number: ${{TRIGGERED_BUILD_NUMBER_PULP_SMASH_RUNNER}}
             flatten: true
         - shell: |
-            if [ "${{PULP_VERSION}}" = "2.8" ]; then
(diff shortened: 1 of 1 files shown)


seandst:
LGTM, but wouldn't mind a comment above conditionals like that talking about why we only want to run coverage on 2.8, and then not on rhel6.

ci/jobs/pulp-dev.yaml
```diff
@@ -63,7 +63,10 @@
             build-number: ${{TRIGGERED_BUILD_NUMBER_PULP_SMASH_RUNNER}}
             flatten: true
         - shell: |
-            if [ "${{PULP_VERSION}}" = "2.8" ]; then
+            # Coverage report is introduced on Pulp 2.8 and requires Python
```
elyezer: @seandst what do you think?

PR #124, opened 2016-04-07
Update the plugins unittest template name

elyezer:
Update the plugins unittest template name on unittests project in order to
match the right job.


elyezer:
Never mind about this, just realized the name was the group of jobs.

PR #128, opened 2016-04-13
Use Ansible's ``file`` module

jeremycline:
Previously /etc/pki/CA/index.txt was being created with ``command``. This
led to problems (and was lazy of me to use in the first place) on EL6
because ``touch`` is ``/bin/touch`` rather than ``/usr/bin/touch``.


elyezer:
:+1:

PR #131, opened 2016-04-13
determine which PKG_MGR works to select one

seandst:
I took another stab at generating nodepool images in CIOS, and found
that PKG_MGR wasn't getting set because apparently "which" isn't
installed in the f22 images there. I tweaked it to be what's here, which
is "use dnf if you can, otherwise use yum".

### DIFF
--- a/ci/nodepool/scripts/bootstrap.sh
+++ b/ci/nodepool/scripts/bootstrap.sh
@@ -6,7 +6,14 @@ export DISTRIBUTION=$(python -c "import platform, sys
 sys.stdout.write(platform.dist()[0])")
 export DISTRIBUTION_MAJOR_VERSION=$(python -c "import platform, sys
 sys.stdout.write(platform.dist()[1].split('.')[0])")
-export PKG_MGR=$(which dnf 2>/dev/null || which yum 2>/dev/null)
+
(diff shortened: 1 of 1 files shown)


ci/nodepool/scripts/bootstrap.sh
```diff
@@ -6,7 +6,14 @@ export DISTRIBUTION=$(python -c "import platform, sys
 sys.stdout.write(platform.dist()[0])")
 export DISTRIBUTION_MAJOR_VERSION=$(python -c "import platform, sys
 sys.stdout.write(platform.dist()[1].split('.')[0])")
-export PKG_MGR=$(which dnf 2>/dev/null || which yum 2>/dev/null)
+
+# use dnf if you can, otherwise use yum
+if dnf --version; then
```
elyezer: `dnf --version &>/dev/null`?

seandst: I'm happy to redirect it, but the silence of the line I'm replacing made it a little difficult to debug. Being able to see that dnf either printed a version or was not found makes it quite clear in the image building log why dnf was (or was not) chosen.

elyezer: @seandst fair enough

PR #134, opened 2016-04-15
Fix SSL errors on pulp-smash-runner job

elyezer:
Pass the hostname of the Pulp server machine instead of the IP address and use
the certificate when doing HTTP requests trough python-requests.


Ichimonji10:
:+1:

PR #138, opened 2016-04-19
Comment Sonar configuration for Pulp automation

elyezer:
Comment the Sonar build step in order to no run it. When the proper
configuration the configuration should be updated and this code block
uncommented.


dkliban:
Why is sonar not ready? What needs to be done for it to be ready?

elyezer:
@dkliban there are more required configuration in order to make it work. I sent a sample coverage.xml grabbed from a Pulp Smash run against Pulp 2.8 on RHEL 7 to Greg and he will create a sample job to check all configuration that is needed. That job will serve as an example for configuring the others.

PR #139, opened 2016-04-20
Update puppet installation for prepare_node.sh

elyezer:
Update the puppet repository URL to new format where the package collection is
specified. And use the changed path for the puppet binaries, for more
information check the documentation [1].

[1] https://docs.puppet.com/puppet/4.4/reference/whered_it_go.html


elyezer:
With these changes I was able to update the images for all RHEL based labels.

PR #148, opened 2016-05-05
Remove all non-RPM content.

rbarlow:


elyezer:
I see that this is now added to the f23 branch. Can you explain a little bit what is the plan for it?

rbarlow:
On Thu, May 05, 2016 at 03:37:03PM -0700, Elyézer Rezende wrote:
> I see that this is now added to the f23 branch. Can you explain a little bit what is the plan for it?

Sure! For one, my grand plan for this repository is actually to delete
it (muahahaha) as the majority of the code in here is not related to
packaging. Instead, I think we need two repositories: devel and
packaging.

devel: This repository will contain all the CI stuff, and also all the
Vagrant and other playpen stuff.

packaging: It's silly that this repository is called
pulp/pulp_packaging, especially given that it doesn't contain our
packaging. Since I'm doing new work, it makes sense to just create a
packaging repository. In fact, I may just delete this PR and do that
now.

The new work I am doing is to get our packaging out of the pulp
repository and into a packaging repository. There will be a branch for
each distro that we support (like f23). If we do this and create that
devel repository, the pulp repository should finally be free of all
non-upstream code and silly tags.

-- 
Randy Barlow
irc:   bowlofeggs

rbarlow:
I created a new repository to hold the f23 stuff, so I'm closing this. See https://github.com/pulp/packaging

PR #149, opened 2016-05-06
Trim down an if/else block

Ichimonji10:
Use the `elif` keyword to make a block of code more compact.


Ichimonji10:
```sh
if [ 1 -eq 2 ]; then
  echo foo
elif [ 1 -eq 1 ]; then
  echo bar
fi
```

Ichimonji10:
Thank you, @elyezer.

PR #154, opened 2016-05-10
Updates master config to use master branch for crane

bmbouter:
Since it uses master the parent_branch can be removed. Also the
usage of master means the version is changing to 2.0.1-0.1beta

re #950


bmbouter:
Closing because we're moving back the docs changes in crane to `2.0-release` since they are only docs.

PR #157, opened 2016-05-12
Add pulp-upgrade jobs

elyezer:
The Pulp Upgrade jobs ensures that Pulp can be upgraded from a given X.Y
version to a X.Y+1 version on the Pulp supported operating systems.

### DIFF
--- a/ci/jobs/pulp-upgrade.yaml
+++ b/ci/jobs/pulp-upgrade.yaml
@@ -1,0 +1,166 @@
+# This job test Pulp upgrading from a X.Y to a X.Y+1 version on many Operating
+# Systems.
+
+- job-template:
+    name: 'pulp-upgrade-{pulp_version}-{upgrade_pulp_version}-{os}'
+    node: '{os}-vanilla-np'
+    scm:
+        - pulp-packaging-github
+    wrappers:
+        - config-file-provider:
+            files:
+                - file-id: rhn_credentials
+                  variable: RHN_CREDENTIALS
+        - credentials-binding:
+            - file:
+                credential-id: c7fb90a7-154c-45d6-b820-a742cf300615
+                variable: CDN_CERTIFICATES
+        - inject:
+            properties-content: |
+                OS={os}
+                PULP_PASSWORD=admin
+                PULP_USER=admin
+                PULP_VERSION={pulp_version}
+                UPGRADE_PULP_VERSION={upgrade_pulp_version}
+    builders:
+        - shell: |
+            sudo yum install -y git ansible libselinux-python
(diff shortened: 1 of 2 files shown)


ci/jobs/pulp-upgrade.yaml
```diff
@@ -0,0 +1,149 @@
...
+                --feed https://cdn.redhat.com/content/dist/rhel/rhui/server/7/7.2/x86_64/os/ \
+                --feed-ca-cert cdn/cdn.redhat.com-chain.crt \
+                --feed-cert cdn/914f702153514b06c1ef279db9dcadce.crt \
+                --feed-key cdn/914f702153514b06c1ef279db9dcadce.key \
+                --repo-id rhel7 \
+                --skip erratum
+            pulp-admin rpm repo sync run --repo-id rhel7 >/dev/null
+
+            pulp-admin logout
+        - shell: |
```
elyezer: Instead of one big shell block I am splitting in some blocks this will allow further improvements like splitting the scripts in different files based on its role, for example, install Pulp, populate Pulp, upgrade Pulp, check populated content, so on and so forth.

This PR is intended to release the first version which is already passing on a RHEL7 system. Further improvements are about to come.

elyezer:
Talked with Preethi on IRC and we decided to trigger these jobs weekly. I will add a trigger job to facilitate it and also allow easy manual trigger when needed.

preethit:
+1 Thank you @elyezer for this and the jenkins job. It will make my life 100 times easier :)

ci/jobs/pulp-upgrade.yaml
```diff
@@ -0,0 +1,164 @@
...
+
+            sudo yum repolist
+
+            sudo yum -y update
+            sudo -u apache pulp-manage-db
+            sudo systemctl daemon-reload
+
+            for service in httpd pulp_workers pulp_celerybeat pulp_resource_manager; do
+                sudo systemctl restart "${{service}}"
+            done
```
Ichimonji10: This loop can be simplified to:

    systemctl restart httpd pulp_workers pulp_celerybeat pulp_resource_manager

The more concise syntax is actually preferable, because systemctl can restart the units in parallel, and can take care of any ordering dependencies if any are present.

Ichimonji10:
This looks fine to me. If it works, let's ship it, and iterate over time.

ci/jobs/pulp-upgrade.yaml
elyezer: @Ichimonji10 thank you very much I will update the PR and merge it.

elyezer:
I am merging this PR and further improvements will be sent in  future PRs

PR #158, opened 2016-05-13
Install patch on pulp-fixtures-publisher job

elyezer:
Make sure that the patch command is available when running make fixtures.


Ichimonji10:
This is definitely needed. ACK.

Ichimonji10:
Specifically, this is needed by the `fixtures/rpm-*` make targets advertised by Pulp Fixtures. Those make targets call a script that patches an XML file.

PR #160, opened 2016-05-13
Add jenkins-sudoers

elyezer:
The file is used by the nodepool image bootstrap.sh script


Ichimonji10:
For reference: https://github.com/pulp/pulp_packaging/blob/master/ci/nodepool/scripts/bootstrap.sh

Ichimonji10:
The sudo syntax looks fine, and I see the reference to this script in `bootstrap.sh`. Ship it.

PR #161, opened 2016-05-13
Use sudo to change selinux configuration

elyezer:
When preparing the node, nodepool does not use the root user and because that
sudo should be used to make sure the edit selinux configuration.


Ichimonji10:
Whoops! Easy fix. Let's ship it. :smile:

PR #166, opened 2016-05-19
Fix escaping issue on pulp-dev jobs

elyezer:


Ichimonji10:
Looks great.

```
[ichimonji10@beech:~]$ echo "foo\\bar"
foo\bar
[ichimonji10@beech:~]$ echo -e "foo\\nbar"
foo
bar
```

PR #171, opened 2016-05-31
Push the version bump commit on the branch where it took place.

seandst:
Without this change, the update-version-and-merge-forward script
updates the version on e.g. 2.8-dev locally, doesn't push it, merges
2.8-dev to master, and then pushes master, requiring a manual push
of the 2.8-dev version bump commit. With this change, the 2.8-dev
version bump is pushed before being merged to master and pushed.


elyezer:
Please rebase since I have merged my PR for the job-ownership.

PR #173, opened 2016-06-01
Two fixes for docs-ownership role

bmbouter:
* Adds ttereshc as the co-owner
* Renames the job as discussed on IRC


bmbouter:
@elyezer thanks for the review.

PR #174, opened 2016-06-02
Add pulp-installer job

elyezer:
This job installs Pulp on the machine specified by its HOSTNAME parameter. Will
be used the same playbook that Pulp Automation uses, this means that the
machine can be used to run Pulp Smash tests.


preethit:
Looks good to me!

PR #175, opened 2016-06-02
build promotion scripts

seandst:
The new "promote-build" script is triggered after the repo build automation
jobs, and blocks until the pulp-(version)-dev builds are complete. If they
succeed, the repo is published to the tested repo location specified in
the release config named on the command line.

fixes #1927
https://pulp.plan.io/issues/1927


ci/config/releases/2.8-dev.yaml
```diff
@@ -1,5 +1,6 @@
 koji-target-prefix: pulp-2.8
 rsync-target-dir: 2.8/stage
+rsync-tested-dir: 2.8/dev
```
seandst: Assuming this approach is approved, I'll put together a docs patch to document this new field.

ci/jobs/projects.yaml
```diff
@@ -11,6 +11,16 @@
...
 
 - project:
+    name: build-automation-promote
+    jobs:
+     - 'build-automation-promote-{release_config}'
+    release_config:
+      - 2.8-dev:
+         version: 2.8
+      - master:
+         version: 2.9
```
seandst: The way versions are inserted here is a little silly. I wonder if we might want to just stash this value in the release config or otherwise derive it, somehow. I'm content to let it work this way for now and improve it later.

ci/jobs/build-automation.yaml
```diff
@@ -62,15 +62,66 @@
...
+            if [[ $DRY_RUN = "true" ]]; then promote_opts="$promote_opts --dry-run"; fi
+
+            ./promote-build.py $promote_opts --job-prefix pulp-{version}-dev {release_config}
+    publishers:
+      # - mark-node-offline is omitted: you can't have more than one groovy postbuild step, and
+      #   groovy is used here to mark the build as unstable when promotion is blocked
+      - groovy-postbuild: |
+          if(manager.logContains(".*Promotion blocked.*")) {{ manager.buildUnstable() }}          
+          manager.build.getBuiltOn().toComputer().setTemporarilyOffline(true)
+      - irc-notify-all-summary
```
elyezer: This is missing the ownership email macro.

ci/jobs/macros.yaml
```diff
@@ -5,6 +5,15 @@
 # they're easy to find.
 ###
 
+# Inject the PYTHONUNBUFFERED env var into builds using a parameter
+- parameter:
+    name: python-unbuffered
+    parameters:
+      - string:
```
elyezer: Would not be better to use a boolean? It will end up with true or false as a string on the variable.

elyezer:
Looks good to me, I left some suggestions feel free to incorporate them or not and merge.

ci/jobs/macros.yaml
seandst: I'll add a comment. PYTHONUNBUFFERED works if it has any value, so setting it to boolean and leaving it unchecked will set it to "false", which will turn PYTHONUNBUFFERED on. The empty string is the only way to turn it off.

elyezer: Thanks for the clarification

PR #176, opened 2016-06-02
Improve Pulp Upgrade jobs

elyezer:
Fix a escaping issue when checking for Pulp 2.8 and not direct repository
syncing to `/dev/null` in order to make easy to debug when it fails.


Ichimonji10:
ACK

PR #179, opened 2016-06-06
possible rhel7-vanilla-np image fix

seandst:


seandst:
I started messing around in the grub rescue shell, and discovered that manually entering the grub config lines with "linux/initrd" instead of "linux16/initrd16" got the images to successfully boot. Booting with "linux16/initrd16" causes grub to report an out of memory error, so the simple fix was to remove the lines that trigger the switch to the *16 variants with sed. I'm assuming that there's something weird going on with the splash screen based on the bug report I mentioned in the comment, but I haven't dug in to find out exactly what. It's possible that just removing plymouth might also make things work, for example.

I don't think this is the best solution to the problem, but if it gets the images building every day without our intervention it might be good enough.

elyezer:
Let's try it.

LGTM

PR #180, opened 2016-06-07
Improve pulp upgrade automation

elyezer:
Add jobs to test from current Y stream stable build to the current Y nightly
build. Make the job template generic in order to be able to upgrade from
any build (beta, nightly and stable) to any build by updating the repository
properly.

Add permutations that specify not just the Pulp version but the build as well,
this is going to allow more permutations and have more control and information
from which Pulp version and build are being upgrading from and to.

Fix missing `sudo` when restarting services after the upgrade.

### DIFF
--- a/ci/jobs/projects.yaml
+++ b/ci/jobs/projects.yaml
@@ -63,8 +63,11 @@
         - f23
         - rhel6
         - rhel7
-    pulp_version: 2.7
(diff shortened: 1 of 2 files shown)


preethit:
Looks good!

Ichimonji10:
This all looks sane to me.

PR #182, opened 2016-06-09
Manually remove python-semantic-version

elyezer:
Uninstall python-semantic-version before upgrading the Pulp packages on Fedora
22 due to issue when updating it.


Ichimonji10:
ACK.

The `--nodeps` option makes the fix seem hacky, but if it's recommended by the Pulp documentation and works, then that's good enough.

elyezer:
@Ichimonji10 agreed. But let's just do that for F22 and soon we should drop it and focus on f23 and 24 which don't requires the hacky fix.

Ichimonji10:
:+1:

PR #185, opened 2016-06-15
Wraps pulp-dev permutations into single quotes

elyezer:
This makes jenkins-jobs-builder generate properly the permutations. For
example, for Pulp 2.10 it was generating 2.1 without the quotes.

### DIFF
--- a/ci/jobs/projects.yaml
+++ b/ci/jobs/projects.yaml
@@ -32,14 +32,14 @@
 - project:
     name: pulp-dev
     os:
-        - f22
-        - f23
-        - rhel6
-        - rhel7
+        - 'f22'
+        - 'f23'
+        - 'rhel6'
+        - 'rhel7'
     pulp_version:
-        - 2.8
-        - 2.9
-        - 2.10:
+        - '2.8'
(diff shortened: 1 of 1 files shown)


ci/jobs/projects.yaml
```diff
@@ -32,14 +32,14 @@
...
+        - 'f23'
+        - 'rhel6'
+        - 'rhel7'
     pulp_version:
-        - 2.8
-        - 2.9
-        - 2.10:
+        - '2.8'
+        - '2.9'
+        - '2.10':
```
seandst: These quotes make sense since yaml will likely try to cast "2.8" as a float, for example.

ci/jobs/projects.yaml
```diff
@@ -32,14 +32,14 @@
...
     name: pulp-dev
     os:
-        - f22
-        - f23
-        - rhel6
-        - rhel7
+        - 'f22'
+        - 'f23'
+        - 'rhel6'
+        - 'rhel7'
```
elyezer: not necessary, but is better to have them here.

PR #186, opened 2016-06-15
Make pulp-smash-runner fail properly

elyezer:
The test runner command returns a non-zero return code whenever at least one
failure happens and jenkins will understand that as being a failure, which will
then send an email to the owners.

In order to avoid the noise for test failures, make the pulp-smash-runner job
fail and notify owners if something serious happens.


Ichimonji10:
LGTM

PR #188, opened 2016-06-17
stop spamming devs with inappropriate promote job failures

seandst:
re #2022
https://pulp.plan.io/issues/2022


seandst:
I've modified this slightly to change how the promotion script exits, which might actually fix #2022. I'm going to test it and see.

PR #189, opened 2016-06-17
Adds a docs builder Jenkins job

bmbouter:
This job nightly build one sphinx project which includes
platform and all plugins develped by the core dev team.

This job uses the docs-ownership so it has owners. It also
uses the e-mail notify macro for owner notification on
failure.

The version of each plugin is specified in the releases yaml
files used by the build infrastructure. Each repo is cloned
freshly with each build.

This job comes with additional rst files which are merged in
prior to the build. This additional content organizes all of
the other content which was not designed to be shown as one
site.

It also deploys a robots.txt for the root of the docs site.

re #950

### DIFF
--- a/ci/docs-builder.py
+++ b/ci/docs-builder.py
@@ -1,0 +1,171 @@
+#!/usr/bin/env python2
+
+import argparse
+import subprocess
+import sys
+import os
+from shutil import copyfile
+
+import yaml
+
+from lib import builder
+from lib.builder import WORKING_DIR
+
+
+LATEST = '2.8'
+
+USERNAME = '57600fb10c1e664383000229'
+HOSTNAME = 'docs-pulp.rhcloud.com'
+
+SITE_ROOT = '~/app-root/repo/diy/'
+
+
+def get_components(configuration):
+    # Get the components from the yaml file
+    repos = configuration['repositories']
+    for component in repos:
+        yield component
+
+
+def load_config(config_name):
+    # Get the config
+    config_file = os.path.join(os.path.dirname(__file__),
+                               'config', 'releases', '%s.yaml' % config_name)
+    if not os.path.exists(config_file):
+        print "Error: %s not found. " % config_file
+        sys.exit(1)
+    with open(config_file, 'r') as config_handle:
+        config = yaml.safe_load(config_handle)
+    return config
+
+def main():
+    # Parse the args
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--release", required=True, help="Build the docs for a given release.")
+    opts = parser.parse_args()
+
+    configuration = load_config(opts.release)
+
+    # Get platform build version
+    repo_list = configuration['repositories']
+    try:
+        pulp_dict = filter(lambda x: x['name'] == 'pulp', repo_list)[0]
+    except IndexError:
+        raise RuntimeError("config file does not have an entry for 'pulp'")
+    version = pulp_dict['version']
+
+    if version.endswith('alpha'):
+        build_type = 'nightly'
+    elif version.endswith('beta'):
+        build_type = 'dev'
+    elif version.endswith('rc'):
+        build_type = 'dev'
+    else:
+        build_type = 'ga'
+
+    x_y_version = '.'.join(version.split('.')[:2])
+
+    builder.ensure_dir(WORKING_DIR, clean=True)
+
+    print "Getting git repos"
+    for component in get_components(configuration):
+        #clone the repos
+        branch_name = component['git_branch']
+        print "Cloning from github: %s" % component.get('git_url')
+        print "Switching to branch %s" % branch_name
+        clone_command = ['git', 'clone', component.get('git_url'), '--branch', branch_name]
+        exit_code = subprocess.call(clone_command, cwd=WORKING_DIR)
+        if exit_code != 0:
+            raise RuntimeError('An error occurred while cloning the repo.')
+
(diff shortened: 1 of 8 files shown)


ci/docs-builder.py
```diff
@@ -0,0 +1,158 @@
...
+    make_command = ['make', 'html', 'SPHINXOPTS=-Wn']
+    exit_code = subprocess.call(make_command, cwd=docs_directory)
+    if exit_code != 0:
+        raise RuntimeError('An error occurred while building the docs.')
+
+    # rsync the docs to the root if it's GA of latest
+    if build_type == 'ga' and x_y_version == LATEST:
+        local_path_arg = docs_directory + os.sep + '_build' + os.sep + 'html' + os.sep
+        remote_path_arg = '57600fb10c1e664383000229@docs-pulp.rhcloud.com:~/app-root/repo/diy/'
+        rsync_command = ['rsync', '-avzh', '--delete', '--exclude', 'en', local_path_arg, remote_path_arg]
```
seandst: `rsync_command` appears to get created and run a lot, and should probably also be generalized in a function.

ci/docs-builder.py
```diff
@@ -0,0 +1,158 @@
...
+    docs_directory = WORKING_DIR + os.sep + 'pulp' + os.sep + 'docs'
+    make_command = ['make', 'html', 'SPHINXOPTS=-Wn']
+    exit_code = subprocess.call(make_command, cwd=docs_directory)
+    if exit_code != 0:
+        raise RuntimeError('An error occurred while building the docs.')
+
+    # rsync the docs to the root if it's GA of latest
+    if build_type == 'ga' and x_y_version == LATEST:
+        local_path_arg = docs_directory + os.sep + '_build' + os.sep + 'html' + os.sep
+        remote_path_arg = '57600fb10c1e664383000229@docs-pulp.rhcloud.com:~/app-root/repo/diy/'
```
seandst: This hostname appears multiple times in the script and looks like something that might change in the future; generating paths to the rsync remote should probably be generalized, and the hostname for the rsync remote should probably be a top-level constant, or otherwise called out as a configuration value.

ci/docs-builder.py
```diff
@@ -0,0 +1,158 @@
...
+    pulp_index_rst = WORKING_DIR + os.sep + 'pulp' + os.sep + 'docs' + os.sep + 'index.rst'
+    copyfile(src_path, pulp_index_rst)
+
+    # copy in the plugin_index.rst file
+    plugin_index_rst = plugins_dir + os.sep + 'index.rst'
+    copyfile('docs/plugin_index.rst', plugin_index_rst)
+
+    # copy in the all_content_index.rst file
+    all_content_index_rst = WORKING_DIR + os.sep + 'pulp' + os.sep + 'docs' + os.sep + 'all_content_index.rst'
+    copyfile('docs/all_content_index.rst', all_content_index_rst)
```
seandst: I'd be sorely tempted to crush all these copyfiles down into a loop, like with `destination: source` pairs in a dict, but appreciate the readability of having them all explicitly written out, particularly in a piece of build automation, since it makes it easier to see exactly what's happening.

ci/docs-builder.py
```diff
@@ -0,0 +1,158 @@
...
+
+        src = WORKING_DIR + os.sep + component['name'] + os.sep + 'docs'
+        dst = plugins_dir + os.sep + component['name']
+        os.symlink(src, dst)
+
+    # copy in the pulp_index.rst file
+    if x_y_version == '2.8' and build_type == 'ga':
+        src_path = 'docs/pulp_index_2_8.rst'
+    else:
+        src_path = 'docs/pulp_index.rst'
```
seandst: My eyeball-differ doesn't see the difference between these two files. Can a comment be added to explain why we want to source the index from a different file based on version?

ci/docs-builder.py
```diff
@@ -0,0 +1,158 @@
...
+            raise RuntimeError('An error occurred while cloning the repo.')
+
+    plugins_dir = WORKING_DIR + os.sep + 'pulp' + os.sep + 'docs' + os.sep + 'plugins'
+    builder.ensure_dir(plugins_dir)
+
+    for component in get_components(configuration):
+        if component['name'] == 'pulp':
+            continue
+
+        src = WORKING_DIR + os.sep + component['name'] + os.sep + 'docs'
```
seandst: Does `os.path.join` not work here (and all the other places where `+ os.sep +` occurs)? If not, I think even `os.sep.join([WORKING_DIR, component['name'], 'docs'])` would lower the amount of concatenation

seandst:
I dropped a lot of comments that can all be disregarded as needless for this sort of automation script (but I'm really rooting for `os.path.join`).

ci/docs-builder.py
bmbouter: A great idea. I've updated it's usage throughout.

ci/docs-builder.py
bmbouter: Oh yes the difference is subtle having to do with a structural change between 2.8.4 and 2.8.5+. I'll add more comments on it now.

ci/docs-builder.py
bmbouter: I'll leave this alone for readability

ci/docs-builder.py
bmbouter: This is a good refactor. I'm doing this.

ci/docs-builder.py
bmbouter: I'm skipping this one because it's made 3 times and each one is unique.

PR #190, opened 2016-06-20
fix deprecations and cobertura type problems

seandst:
The cobertura plugin wants ints, not floats, for coverage values. This was
causing the "Error in request. Possibly authentication failed [500]: Server Error"
errors that we've been seeing.

groovy-postbuild was changed to want the groovy script in a "script"
dict config key, so I did that where needed to make that deprecation
warning go away

ssh-agent-credentials wants an array of users instead of a single user
param now in its config dict, so I've updated jobs using that wrapper
to be configured the new way. This has also become a macro, because
if I'm going to fix it the same way in four places I might as well
make it a macro.

### DIFF
--- a/ci/jobs/build-automation.yaml
+++ b/ci/jobs/build-automation.yaml
@@ -25,8 +25,7 @@
           description: "Perform a release build, defaults to 'false'."
       - python-unbuffered
     wrappers:
-        - ssh-agent-credentials:
-            user: '044c0620-d67e-4172-9814-dc49e443e7b6'
+        - jenkins-ssh-credentials
         - credentials-binding:
             - zip-file:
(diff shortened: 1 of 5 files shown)


ci/jobs/build-automation.yaml
```diff
@@ -25,8 +25,7 @@
           description: "Perform a release build, defaults to 'false'."
       - python-unbuffered
     wrappers:
-        - ssh-agent-credentials:
-            user: '044c0620-d67e-4172-9814-dc49e443e7b6'
+        - jenkins-ssh-credentials
```
elyezer: :+1:

elyezer:
This looks great, thanks for updating all the configurations for the ssh-agent.

LGTM

PR #192, opened 2016-06-20
Adds analytics tracking to docs.pulpproject.org

bmbouter:
re #950

### DIFF
--- a/ci/docs-builder.py
+++ b/ci/docs-builder.py
@@ -109,6 +109,14 @@ def main():
     all_content_index_rst = os.sep.join([WORKING_DIR, 'pulp', 'docs', 'all_content_index.rst'])
     copyfile('docs/all_content_index.rst', all_content_index_rst)
 
+    # make the _templates dir
+    layout_dir = os.sep.join([WORKING_DIR, 'pulp', 'docs', '_templates'])
+    os.makedirs(layout_dir)
+
+    # copy in the layout.html file for analytics
+    layout_html_path = os.sep.join([WORKING_DIR, 'pulp', 'docs', '_templates', 'layout.html'])
+    copyfile('docs/layout.html', layout_html_path)
+
     # build the docs via the Pulp project itself
     print "Building the docs"
     docs_directory = os.sep.join([WORKING_DIR, 'pulp', 'docs'])
--- a/ci/docs/layout.html
+++ b/ci/docs/layout.html
@@ -1,0 +1,15 @@
+{% extends "!layout.html" %}
+
+{% block footer %}
+{{ super() }}
+<script>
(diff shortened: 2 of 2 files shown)


ci/docs-builder.py
```diff
@@ -109,6 +109,14 @@ def main():
     all_content_index_rst = os.sep.join([WORKING_DIR, 'pulp', 'docs', 'all_content_index.rst'])
     copyfile('docs/all_content_index.rst', all_content_index_rst)
 
+    # make the _templates dir
+    layout_dir = os.sep.join([WORKING_DIR, 'pulp', 'docs', '_templates'])
```
elyezer: Why not use `os.path.join`?

elyezer:
Left a non-block comment.

It looks good to me. My only concern is that will Pulp docs users like to be tracked by Google Analytics? I am not too worried about that, but I saw some conversations about using Analytics and wanted to make sure.

ci/docs-builder.py
bmbouter: This was a good question. I hadn't thought much about it, but after reading some I don't want the behavior that a path component will be replaced by an absolute path. I don't want that behavior because in several places paths are meant for a remote machine which has a different absolute path. `os.sep` doesn't do this.

bmbouter:
@elyezer Thanks for the review. I also have some concern that some users may not prefer this tracking. The data would be valuable for the community, so I lean slightly in favor of adding it. Also users can use the analytics opt-out plugins as a way to not be included. It's a good question, I'm glad you asked it.

ci/docs-builder.py
elyezer: In order to get the abspath then the os.path.abspath should be called, otherwise the behavior of os.path.join and os.sep.join are the same:

```python
>>> import os
>>> os.path.join('a', 'b')
'a/b'
>>> os.sep.join(['a', 'b'])
'a/b'
```

The difference between both is that the former is a helper module and the later is a plain string:

```python
>>> os.path
<module 'posixpath' from '/home/elyezer/.virtualenvs/pulp-smash/lib64/python3.4/posixpath.py'>
>>> os.sep
'/'
```

With the above said, they will generate the same result and both of them can be used. Personally I like `os.path` more because its meaning is more clear to me. This is not required to be changed though.

Ichimonji10: `os.path.join` can better handle mistakes made by users:

```python
>>> import os
>>> os.path.join('foo', 'bar')
'foo/bar'
>>> os.path.join('foo/', 'bar')
'foo/bar'
>>> os.sep.join(['foo', 'bar'])
'foo/bar'
>>> os.sep.join(['foo/', 'bar'])
'foo//bar'
```

bmbouter: Oh very good. Thanks for the info.

PR #195, opened 2016-06-23
Add upgrade jobs for Pulp 2.7 stable to 2.8 beta

elyezer:
Also exclude unwanted permutations like Pulp 2.7 stable on Fedora 23 because
there are no packages released also upgrade from Pulp 2.8 stable to 2.9 beta.


preethit:
LGTM!

PR #198, opened 2016-06-27
Add merge-forward-only option to update-and-merge-forward script

seandst:
It would probably be better to have a generic "merge forward" script
that all devs could use. In lieu of that, it was trivial to tweak
this script to make it correctly merge branches forward based on
a release config.


elyezer:
I can't review the new option requirement but the code looks good to me.

seandst:
I needed this yesterday and actually ran it, so I know it works, which is nice. :)

PR #203, opened 2016-07-11
Fix up version so we promote '2.10', not '2.1'

seandst:
Versions need to be strings, not floats, to keep the trailing zero


ci/jobs/projects.yaml
```diff
@@ -13,11 +13,11 @@
      - 'build-automation-promote-{release_config}'
     release_config:
       - 2.8-dev:
-         version: 2.8
+         version: '2.8'
```
elyezer: :+1:

PR #206, opened 2016-07-12
build-all can skip scratch for release builds, promote regex fix

seandst:
- My release process includes a manual scratch build step, so doing a
  scratch build in the release step is redundant. Note that the scratch
  build can only be skipped for a release build. (build-all.py)
- A regex in lib/promote.py was unanchored, and inappropriately
  interpreting '2.9.0-releasenotes' as '9.0-release'. A terminal
  string anchor fixes this issue.


elyezer:
I can't review the procedure but the code looks good.

PR #208, opened 2016-07-15
Correctly compare Pulp version strings

Ichimonji10:
When comparing two version strings, it's invalid to do this:

    version | float >= 2.8

The problem is that `version` may be a value such as '2.10', which is
cast to the float value of 2.1. The `version_compare` filter should be
used instead.

Thanks to @elyezer for help with tracking down this issue.

Fix: https://pulp.plan.io/issues/2080


Ichimonji10:
I apologize for goofing up, @dkliban. Using `version_compare` *is* the right fix.

PR #209, opened 2016-07-18
Remove old JJB build configs

seandst:
https://pulp.plan.io/issues/1968
fixes #1968


elyezer:
Good to see f22 dropped and some older versions of Pulp as well.

PR #213, opened 2016-08-09
Pull distribution info from /etc/os-release

seandst:
f24 builds were failing because, for whatever reason, the package
module was returning empty strings for dist info. Since we aren't
building for el5 anymore, we can read /etc/os-release to get the
dist information for setting DISTRIBUTION and _MAJOR_VERSION.

### DIFF
--- a/ci/nodepool/scripts/bootstrap.sh
+++ b/ci/nodepool/scripts/bootstrap.sh
@@ -1,11 +1,32 @@
 #!/bin/bash
+# This script is meant to be sourced, not executed directly, and shouldn't do
+# anything more than determine some system basics and get the jenkins user and
+# dependencies installed.
 set -ex
 
-# Collect some facts
-export DISTRIBUTION=$(python -c "import platform, sys
-sys.stdout.write(platform.dist()[0])")
-export DISTRIBUTION_MAJOR_VERSION=$(python -c "import platform, sys
-sys.stdout.write(platform.dist()[1].split('.')[0])")
+# This file doesn't exist on el5, but we don't test on el5 anymore,
+# so that's probably(?) fine. If needed, we could add special handling for el5
+. /etc/os-release
+
+if [ -z $ID ] || [ -z $VERSION_ID ]; then
+   echo "ID and VERSION_ID not set in /etc/os-release"
+   return 1
+fi
+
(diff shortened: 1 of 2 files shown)


ci/nodepool/scripts/prepare_node.sh
```diff
@@ -39,29 +39,19 @@ elif  [ "${DISTRIBUTION}" == "redhat" ] && [ "${DISTRIBUTION_MAJOR_VERSION}" ==
...
 
-if  [ "${DISTRIBUTION}" == "fedora" ] && [ "${DISTRIBUTION_MAJOR_VERSION}" == "23" ]; then
-    PUPPET="puppet"
-    sudo "${PKG_MGR}" install -y puppet
-else
-    PUPPET="/opt/puppetlabs/bin/puppet"
-    sudo "${PKG_MGR}" install -y puppet-agent
-fi
+PUPPET="/opt/puppetlabs/bin/puppet"
+sudo "${PKG_MGR}" install -y puppet-agent
```
seandst: Since I changed it to pull the fedora puppet package based on the fedora major version, we can always install puppet-agent.

If this breaks again, it might be the last straw before we switch to ansible. :)

ci/nodepool/scripts/bootstrap.sh
```diff
@@ -1,11 +1,32 @@
 #!/bin/bash
+# This script is meant to be sourced, not executed directly, and shouldn't do
```
elyezer: Much needed, thanks for adding.

ci/nodepool/scripts/bootstrap.sh
```diff
@@ -1,11 +1,32 @@
...
+# dependencies installed.
 set -ex
 
-# Collect some facts
-export DISTRIBUTION=$(python -c "import platform, sys
-sys.stdout.write(platform.dist()[0])")
-export DISTRIBUTION_MAJOR_VERSION=$(python -c "import platform, sys
-sys.stdout.write(platform.dist()[1].split('.')[0])")
+# This file doesn't exist on el5, but we don't test on el5 anymore,
+# so that's probably(?) fine. If needed, we could add special handling for el5
```
elyezer: Agreed

PR #215, opened 2016-08-11
Add macro to centralize sat6-pulp testing outputs and Break deadloop of broken instance

danuzclaudes:
The OpenStack flavor 'c3.xlarge' is unstable on OS1 platform. Add code to prevent dead loop if an instance was failed to create.


elyezer:
Merged as 6af41afcd3463f00785a1dd0fb0ca6f2568cc631.

PR #217, opened 2016-08-16
Move pulp-upgrade jobs trigger

elyezer:
Move the weekly trigger to each upgrade job since there are some permutations
that are not valid, like Pulp 2.7 on Fedora 23, and that makes the separated
trigger job to fail.


Ichimonji10:
LGTM

elyezer:
Rebased

PR #222, opened 2016-08-26
Update upgrade jobs permutations

elyezer:
Drop all permutations for Pulp 2.7. And add permutations for upgrades starting
from Pulp 2.9 and upgrades to Pulp 2.10 nightly.


preethit:
LGTM

PR #223, opened 2016-08-29
Fallback to Python on nodepool bootstrap script

elyezer:
Fallback to Python to get the DISTRIBUTION and DISTRIBUTION_MAJOR_VERSION when
the `/etc/os-release` file is not present.

### DIFF
--- a/ci/nodepool/scripts/bootstrap.sh
+++ b/ci/nodepool/scripts/bootstrap.sh
@@ -4,29 +4,43 @@
 # dependencies installed.
 set -ex
 
-# This file doesn't exist on el5, but we don't test on el5 anymore,
-# so that's probably(?) fine. If needed, we could add special handling for el5
-. /etc/os-release
+# Use /etc/os-release if available or fallback to Python in order to collect
+# the DISTRIBUTION and DISTRIBUTION_MAJOR_VERSION information.
+if [ -f /etc/os-release ]; then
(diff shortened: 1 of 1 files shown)


ci/nodepool/scripts/bootstrap.sh
```diff
@@ -4,29 +4,39 @@
...
-  DISTRIBUTION=redhat
+    # We only care about the major version, so split on the dots
+    # and only take the first field (7.1 and 7 are stored as 7).
+    DISTRIBUTION_MAJOR_VERSION=$(echo "${VERSION_ID}"|cut -d . -f 1)
 else
-  DISTRIBUTION="$ID"
+    DISTRIBUTION=$(python -c "import platform, sys
+sys.stdout.write(platform.dist()[0])")
+    DISTRIBUTION_MAJOR_VERSION=$(python -c "import platform, sys
+sys.stdout.write(platform.dist()[1].split('.')[0])")
```
seandst: At first glance, this appears to be a more concise and generally better solution, so we probably want a comment explaining that this started to fail in fedora 24, so we switched to parsing /etc/os-release by default, only falling back to using python on older distributions that do not provide /etc/os-release

seandst: ...but we can do that later if you want. I meant to LGTM this since it unblocks el6 :)

PR #224, opened 2016-08-29
Update nodepool config

elyezer:
Update images user attribute to username in order to make the configuration
valid again.


Ichimonji10:
LGTM

PR #225, opened 2016-08-31
Build 3.0-dev docs nightly

seandst:
Since we'll most definitely be making docs changes as part of 3.0-dev,
it seems like a good idea to get the 3.0 nightly builds started up.

### DIFF
--- a/ci/config/releases/3.0-dev.yaml
+++ b/ci/config/releases/3.0-dev.yaml
@@ -22,10 +22,6 @@ repositories:
     git_url: git@github.com:pulp/pulp_docker.git
     git_branch: 3.0-dev
     version: 3.0.0-0.0.alpha
-  - name: crane
-    git_url: git@github.com:pulp/crane.git
-    git_branch: 3.0-dev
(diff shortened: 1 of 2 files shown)


bmbouter:
I'm so glad to see this get enabled. The only issue I can see is that the nightly builder will fail when checking out crane's 3.0-dev branch because it doesn't have one. Should we branch a 3.0-dev from master on pulp/crane or change the branch the docs builder would use [here](https://github.com/pulp/pulp_packaging/blob/master/ci/config/releases/3.0-dev.yaml#L27)?

Since we likely will need the 3.0-dev branch maybe we should do the former. What do you think?

seandst:
I think I'd prefer to just change the crane branch for now, and not get into the question of what we might be doing with crane for 3.0.

bmbouter:
@seandst That sounds fine to me. Can you change it in this PR? I'll go ahead and add the LGTM so once the branch name is changed you can go ahead and merge.

seandst:
Will do, and thanks!

PR #228, opened 2016-09-14
Sync DRPMS and SRPMS when upgrading

elyezer:
Make sure upgrade will not break if DRPMS and SRPMS are synced before upgrading


preethit:
LGTM

Ichimonji10:
LGTM

PR #230, opened 2016-09-19
pulp-smash runner uses python 3 explicitly

seandst:
It was previously using python 2, which no longer works.

### DIFF
--- a/ci/jobs/pulp-dev.yaml
+++ b/ci/jobs/pulp-dev.yaml
@@ -163,7 +163,7 @@
             sudo update-ca-trust
 
             sudo yum -y install python-pip python-virtualenv
-            virtualenv venv
+            virtualenv -p /usr/bin/python3 venv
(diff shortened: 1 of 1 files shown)


ci/jobs/pulp-dev.yaml
```diff
@@ -163,7 +163,7 @@
             sudo update-ca-trust
 
             sudo yum -y install python-pip python-virtualenv
-            virtualenv venv
+            virtualenv -p /usr/bin/python3 venv
```
elyezer: :+1:

Ichimonji10:
FYI, `pyvenv venv` or `python -m venv venv` can also be used to create a virtualenv. The benefit is that `pyvenv` and `python -m venv` ship with Python itself, whereas virtualenv are separate Python packages. In other words, using `pyvenv` or `python -m venv` lets one create a virtualenv with no  dependencies.

PR #231, opened 2016-09-23
Upgrade logs

elyezer:
Create two commits to better separate:

* Externalize the log capturing logic into a macro
* Use the macro to capture the logs on upgrade jobs


Ichimonji10:
LGTM

PR #234, opened 2016-09-28
Customize mirrorlist files

Ichimonji10:
Pulp Fixtures is capable of generating mirrorlist files. The mirrorlist
files contain absolute URLs that, by default, start with a base URL of
http://localhost:8000/…. These URLs are valid when fixtures are served
with `python3 -m http.server`, but not when the fixtures are served with
some other technique. Customize these base URLs.

See: https://github.com/PulpQE/pulp-smash/issues/384


Ichimonji10:
This is designed to work with https://github.com/PulpQE/pulp-fixtures/pull/37.

Ichimonji10:
Rebased.

elyezer:
LGTM

PR #235, opened 2016-10-04
handle NPE in mark-node-offline macro

seandst:


seandst:
Tested in a groovy shell:

```groovy
groovy:000> manager = null
===> null
groovy:000>             try { 
groovy:001>                 manager.build.getBuiltOn().toComputer().setTemporarilyOffline(true)
groovy:002>             } catch(NullPointerException ex) {}
===> null
```

While the failure we'd see happens when `getBuiltOn` returns `null` instead of a `Slave` instance, setting `manager` to `null` simulates the same error condition by attempting to do a lookup on `manager`, which would also cause an NPE.

Compare to a groovysh session without the exception handling:

```groovy
groovy:000> manager = null
===> null
groovy:000> manager.build.getBuiltOn().toComputer().setTemporarilyOffline(true)
ERROR java.lang.NullPointerException:
Cannot get property 'build' on null object
```

PR #236, opened 2016-10-06
{upstream,downstream}-satellite6-upgrade-pulp jobs

elyezer:
I've created 3 different commits to accomplish the following:

* Merge all jobs that was part of the previous downstream-satellite6-upgrade-pulp automation into one job.
* Refactor the merged job into a job-template in order to use the same template for both downstream and upstream satellite.
* The ci/jobs/scripts dir was created to help with the refactor. Since this new dir is a include dir we need to tell JJB that. The third commit create a sample jenkins_jobs.ini file and update the .travis.yml to use it when trying to generate the jobs.

I am currently testing all changes but feel free to leave any review comments. If some change is needed I will update the PR.


ci/jobs/satellite6-upgrade-pulp.yaml
```diff
@@ -0,0 +1,194 @@
...
+            python-version: System-CPython-2.7
+            nature: shell
+            command: |
+                pip install -U pip requests
+        - shining-panda:
+            build-environment: virtualenv
+            name: satellite6-jenkins-trigger
+            python-version: System-CPython-2.7
+            nature: python
+            command: |
```
seandst: I was thinking that just about any time we saw `command: |` followed by a wall of text we'd switch to including from the scripts dir, so we should probably do that here, as well.

ci/jobs/scripts/openstack-upgrade-pulp.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+echo "Start upgrading Pulp on ${INSTANCE_IP}..."
+alias SSH_CLI="ssh -t -o StrictHostKeyChecking=no cloud-user@${INSTANCE_IP}"
```
seandst: Style nit, but I'm not used to seeing UPPER_CASE aliases; these look like envvars to me. Since ssh itself isn't used later on in the script `alias ssh="ssh -t ...` might be good enough.

ci/jobs/scripts/openstack-upgrade-pulp.sh
```diff
@@ -0,0 +1,15 @@
...
+SSH_CLI "sudo katello-service stop"
+SSH_CLI "rpm -qa | grep pulp"
+SSH_CLI "sudo yum-config-manager --add-repo ${PULP_UPGRADE_REPO_URL}"
+SSH_CLI "sudo yum repolist enabled | grep pulp"
+SSH_CLI "sudo yum update -y --nogpgcheck pulp-*"
+SSH_CLI "rpm -qa | grep pulp"
+SSH_CLI "sudo katello-service restart"
+SSH_CLI "sudo -u apache pulp-manage-db"
+UPDATE_PULP_VERSION="$(SSH_CLI rpm -qa | grep pulp-server)"
+UPDATE_PULP_VERSION="$(echo ${UPDATE_PULP_VERSION} | awk -F'-' '{ print $3 }')"
```
seandst: UPDATE_PULP_VERSION="$(SSH_CLI rpm -qa | grep pulp-server) | awk -F'-' '{ print $3 }')" ?

ci/jobs/satellite6-upgrade-pulp.yaml
elyezer: :+1:

ci/jobs/scripts/openstack-upgrade-pulp.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+echo "Start upgrading Pulp on ${INSTANCE_IP}..."
+alias SSH_CLI="ssh -t -o StrictHostKeyChecking=no cloud-user@${INSTANCE_IP}"
+SSH_CLI "exit"
+SSH_CLI "sudo katello-service stop"
+SSH_CLI "rpm -qa | grep pulp"
```
elyezer: :+1:

ci/jobs/scripts/openstack-upgrade-pulp.sh
elyezer: :+1:

elyezer:
Update: rebased

PR #238, opened 2016-10-12
Use local connection when running Ansible

elyezer:


Ichimonji10:
LGTM. It'll get our builds working again.

Do we still need the `ssh-keygen -t rsa -N "" -f pulp_server_key` line?

elyezer:
> Do we still need the ssh-keygen -t rsa -N "" -f pulp_server_key line?

Yes, that is the key that pulp-smash-runner job will use.

I have removed the export ansible env var line, that will not be needed anymore.

Ichimonji10:
Still LGTM.

PR #241, opened 2016-10-19
Update versions before docs build

seandst:
I snuck in a lint commit; the tip commit of this PR is the interesting one.

I'd like @bmbouter's signoff in particular before merging, in case I'm missing a piece of a bigger picture with this.

### DIFF
--- a/ci/docs-builder.py
+++ b/ci/docs-builder.py
@@ -32,7 +32,7 @@ def load_config(config_name):
     config_file = os.path.join(os.path.dirname(__file__),
                                'config', 'releases', '%s.yaml' % config_name)
     if not os.path.exists(config_file):
-        print "Error: %s not found. " % config_file
+        print("Error: %s not found. " % config_file)
         sys.exit(1)
     with open(config_file, 'r') as config_handle:
         config = yaml.safe_load(config_handle)
@@ -111,7 +111,7 @@ def main():
     copyfile('docs/layout.html', layout_html_path)
 
     # build the docs via the Pulp project itself
-    print "Building the docs"
+    print("Building the docs")
     docs_directory = os.sep.join([WORKING_DIR, 'pulp', 'docs'])
     make_command = ['make', 'html', 'SPHINXOPTS=-Wn']
     exit_code = subprocess.call(make_command, cwd=docs_directory)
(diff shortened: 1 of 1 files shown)


elyezer:
The changes looks good to me, but I will @bmbouter to have the final LGTM.

ci/docs-builder.py
```diff
@@ -32,12 +32,21 @@ def load_config(config_name):
...
     with open(config_file, 'r') as config_handle:
         config = yaml.safe_load(config_handle)
     return config
 
+
+def update_version(config_name):
+    script_dir = os.path.abspath(os.path.dirname(__file__))
+    # script does not push to github, but it does check out all the repos we want and
+    # makes sure their version reflect the versions in the build config being used
+    script = os.path.join(script_dir, 'update-version-and-merge-forward.py')
```
bmbouter: What do you think about making the `update-version-and-merge-forward` import without side effects. All we need to do is to put all the module level code in a method like `main()` and then add an `if __name__ == "__main__":` that would call `main()`. Then we could import that from here and call `main()` ourselves.

ci/docs-builder.py
```diff
@@ -67,16 +76,8 @@ def main():
...
 
-    print "Getting git repos"
-    for component in get_components(configuration):
-        #clone the repos
-        branch_name = component['git_branch']
-        print "Cloning from github: %s" % component.get('git_url')
-        print "Switching to branch %s" % branch_name
-        clone_command = ['git', 'clone', component.get('git_url'), '--branch', branch_name]
-        exit_code = subprocess.call(clone_command, cwd=WORKING_DIR)
-        if exit_code != 0:
```
bmbouter: This check needs to be added to the other script. If these subprocess calls fail this job is expected to raise a fatal exception. The other script is missing this in a few places.

ci/docs-builder.py
```diff
@@ -67,16 +76,8 @@ def main():
 
     builder.ensure_dir(WORKING_DIR, clean=True)
```
bmbouter: This is also in the script and is currently redundant. Maybe it will be disincluded when the script is made import side-effect free.

ci/docs-builder.py
seandst: check_call does this. If the update-version scripts fails, this script will fail.

seandst:
I agree with your comments, but consider them enhancements better addressed in subsequent PRs.

ci/docs-builder.py
bmbouter: Oh yes it does, very good.

PR #243, opened 2016-10-31
Push master docs as unversioned "nightly" docs.

seandst:


bmbouter:
@goosemania @asmacdo FYI this publishes the master branch docs to `/en/nightly/` and it runs every night according to its JJB config.

bmbouter:
@seandst Say master branch is 2.12. Do we typically have a release-config for both 2.12 and master? Or do we wait to make the 2.12 until we branch master to 2.12-dev?

seandst:
> Once the first push occurs of /en/nightly/ can an update also be made on http://pulpproject.org/docs/ linking to it please?

Yeah, adding that link to the docs page is my motivation for doing this.

> Say master branch is 2.12. Do we typically have a release-config for both 2.12 and master? Or do we wait to make the 2.12 until we branch master to 2.12-dev?

Excellent question!

No, the release configs (with the exception of the "x.y-build" configs) correspond to branch names. The "master" config should always be the latest and greatest nightly build of the "next" Pulp Y Release, which can be seen when I recently created the 2.11 release configs to correspond wit the new 2.11-dev branch:

https://github.com/pulp/pulp_packaging/commit/1a9340c6c2d97485036aa9e7d016354e260303fa#diff-d326c7d73e18f6abc69820dd8bb35b17

My expectation is that this config layout will continue even after moving to copr-based builds and the corresponding "new-style" release configs in pulp 3.

Also, thanks for looping in the other folks that get emails if I screwed this up. :)

bmbouter:
@seandst Thanks for the info! Your approach for release-configs is clear and looks good.

PR #244, opened 2016-11-01
Adds some logging around query that keeps failing

dkliban:


elyezer:
Changes looks good to me, I would suggest to avoid single use variables but that is not a big deal on this script. I will leave for @bmbouter take a final look.

ci/redmine_bugzilla.py
```diff
@@ -172,6 +183,9 @@ def main():
                         links_issues_record += 'Bugzilla #%s -> Redmine %s, but Redmine %s does ' \
                                                'not link back\n' % (bug.id, issue.id, issue.id)
 
+    end_time = time.time()
+    time_to_query_all = end_time - start_time
+    print 'Total time to query %d redmine issues: %f' % (redmine_issues_looked_up, time_to_query_all)
```
seandst: This is a duplicate block, and should probably be wrapped in a function.

In both cases where this value is printed, it would be helpful to see the unit of time measured.

ci/redmine_bugzilla.py
```diff
@@ -149,12 +150,22 @@ def main():
...
                     add_cc_list_to_bugzilla_bug(bug)
                     issue_id = external_bug['ext_bz_bug_id']
-                    issue = redmine.issue.get(issue_id)
+                    try:
+                        redmine_issues_looked_up += 1
+                        issue = redmine.issue.get(issue_id)
+                    except:
+                        end_time = time.time()
+                        time_to_query_before_failure = start_time - end_time
+                        print 'Total time to query %d redmine issues: %f' %\
```
bmbouter: As part of a separate effort, this script is being converted to Python2.7+ so changes need to be compatible with Python 3. This print statement is not Python3 compatible I believe.

ci/redmine_bugzilla.py
bmbouter: @dkliban Yes please de-duplicate this code with a function.

ci/redmine_bugzilla.py
```diff
@@ -149,12 +150,22 @@ def main():
                         links_issues_record += 'Redmine #%s -> Bugzilla %s, but Bugzilla %s does not ' \
                                            'link back\n' % (issue.id, bug.id, bug.id)
 
+    start_time = time.time()
```
bmbouter: This 1 start_time is being used for multiple measurements. For the total time this makes sense, but it will be wrong for individual queries on L#162. I suggest having a separate start time for each query in L#162 so that the measurement on L#165 will be only for that 1 query. Using separate, and good variable names would be good.

ci/redmine_bugzilla.py
```diff
@@ -149,12 +150,22 @@ def main():
...
+    redmine_issues_looked_up = 0
     for bug in bugzilla_bugs:
         for external_bug in bug.external_bugs:
                 if external_bug['type']['description'] == 'Pulp Redmine':
                     add_cc_list_to_bugzilla_bug(bug)
                     issue_id = external_bug['ext_bz_bug_id']
-                    issue = redmine.issue.get(issue_id)
+                    try:
+                        redmine_issues_looked_up += 1
+                        issue = redmine.issue.get(issue_id)
```
bmbouter: You might consider writing a timing method which times and records timing information and data about the query.

The instead of `issue = redmine.issue.get(issue_id)` could become `issue = run_and_time(redmine.issue.get, [issue_id])`. More generally the signature would be `def run_and_time(callable, args, kwargs)`.

bmbouter: This could be used to record and aggregate information into a simple report that could be printed at the end.

bmbouter: This is just an idea. If this is helpful do it if not ignore it.

ci/redmine_bugzilla.py
elyezer: to use print as function on Python 2.7 need to have the following import:

```python
from __future__ import print_function
```

seandst: Just saw this, and it's not entirely true:
```
$ python2.7
Python 2.7.12 (default, Nov  7 2016, 11:55:55) 
[GCC 6.2.1 20160830] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> print('Hi')
Hi
```

This example explains why...
```
>>> print('Hello' 'World')
HelloWorld
>>> print('Hello', 'World')
('Hello', 'World')
```
The whitespace after `print` doesn't matter, so as long as you aren't using commas to turn what's inside the parenthesis into a tuple, the print statement will handle it just fine.

So as long as you don't get commas involved and turn this into a tuple, I think you're free to `print('foo')` in py2.7.

Ichimonji10: Yes, that's true. Even so, I'd advocate for using `from __future__ import print_function`. One benefit of doing so is that the same function will be used regardless of whether Python 2.7 or 3.x is used to execute the script. In a simple case like `print('Hi')`, it doesn't really matter. In a more complex case like `print('Hi', end='')`, it does matter.

PR #246, opened 2016-11-09
Update docs jobs and builder to install pulp for apidocs

seandst:
re #2327
https://pulp.plan.io/issues/2327

### DIFF
--- a/ci/jobs/docs.yaml
+++ b/ci/jobs/docs.yaml
@@ -34,8 +34,8 @@
     builders:
         - shell: |
             #!/bin/bash
-
-            sudo dnf install graphviz plantuml python-sphinx PyYAML python-sphinx_rtd_theme -y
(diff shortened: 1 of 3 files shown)


ci/jobs/docs.yaml
```diff
@@ -48,9 +48,15 @@
 
             chmod 644 /home/jenkins/.ssh/known_hosts
 
+            # create a virtualenv in which to install packages needed to build docs
+            virtualenv -p /usr/bin/python3 --system-site-packages ~/docs_ve
```
Ichimonji10: As of Python 3.3, the `venv` module is shipped with the standard library. The following commands both create a Python 3 virtualenv:

    python3 -m venv env
    virtualenv -p python3 env

One benefit to using the `venv` module shipped with the standard library is that one less dependency is needed.

PR #247, opened 2016-11-11
Update nodepool config and bootstrap script

elyezer:
Update nodepool.yaml config file to match the current config:

* Add cleanup config on cron section.
* Set pulp-smash-np nodes min-ready to 0, the new schedules is dealing great
  on provisioning nodes and preventing any deadlock.
* Set the max-servers to 18 instead of 19, this will give room for nodepool.
  building images even when operating on full capacity.
* Update rhel7 nodes to use the recent released RHEL 7.3.

Update bootstrap.sh script to not fail when no repo file is present on
/etc/yum.repos.d directory.


elyezer:
Rebased.

PR #248, opened 2016-11-11
Improve ssh-config for pulp-smash runner

elyezer:
Add ControlMaster, ControlPersist and ControlPath options to the
pulp-smash-runner's ssh-config. This will speed up running Pulp Smash tests.


Ichimonji10:
LGTM

PR #249, opened 2016-11-15
pulp3-specific docs index for the docs builder

seandst:


bmbouter:
@seandst Did you run the docs-builder script locally to test this or should I?

bmbouter:
We could just merge and see I guess, it would probably be faster

seandst:
Yeah, it totally failed. I forgot that I had this set up so I could actually test the docs build, so...I'ma do that and fix this up. :)

ci/docs/pulp_index.rst
```diff
@@ -29,8 +29,6 @@ Content
    plugins/index
    dev-guide/index
    dev-guide/contributing/index
-   api/pulp/app/index
-   api/pulp/plugin
```
seandst: These were introduced for pulp3, so they no longer need to be in this index.

seandst:
Updated. I'm not sure how exactly `all_content_index` works, so let me know if what I did was silly and I'll fix it.

bmbouter:
I think you need to git add `all_content_index_pulp3.rst`.

seandst:
Missing file added

PR #250, opened 2016-11-15
Installs gcc and redhat-lsb-core on all nodes for unit test runners

dkliban:


ci/nodepool/scripts/prepare_node.sh
```diff
@@ -14,6 +14,9 @@ source bootstrap.sh
 echo "Performaing a general update"
 sudo "${PKG_MGR}" update -y
 
+echo "Installing redhat-lsb-core and gcc"
+sudo "${PKG_MGR}" install -y redhat-lsb-core gcc
```
seandst: As far as I can tell, redhat-lsb-core isn't available in fedora, so this should be in a conditional.

https://admin.fedoraproject.org/pkgdb/packages/redhat-lsb-core

dkliban: It's available https://apps.fedoraproject.org/packages/redhat-lsb-core

seandst: CURSE YOU, PKGDB.

PR #252, opened 2016-11-20
Updates downstream bug state

bmbouter:
This puts 4 changes in place:

* The downstream bug state is transitioned to POST
  when the upstream bug is at MODIFIED+.

* If the downstream bug is at POST+ and the upstream
  bug is at POST- it fails the job which notifies a
  human.

* A placeholder for a future Foreman integration point
  where downstream transitions will not occur if the
  upstream Foreman bug is not at MODIFIED+.

* Some small fixes where lines were > 100 chars.

https://pulp.plan.io/issues/2351
closes #2351

### DIFF
--- a/ci/redmine_bugzilla.py
+++ b/ci/redmine_bugzilla.py
@@ -17,9 +17,15 @@ the Bugzilla bug. If they are not present add them. REQUIRED_CC is a list
 of Bugzilla user emails stored as a constant in this module.
 
 Requirements:
+There are a few pip installable requirements:
+
+    $ pip install requests python-bugzilla python-redmine urllib3 pyopenssl
+
 The Redmine and Bugzilla credentials are expected from to be read from a
 file in ini format. See the example below. The location of this file is set
 by the REDMINE_BUGZILLA_CONF environment variable.
+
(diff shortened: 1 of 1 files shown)


ci/redmine_bugzilla.py
```diff
@@ -132,22 +142,39 @@ def main():
...
+                                else:
+                                    transition_to_post.append(False)
+                            downstream_POST_plus = ['POST', 'MODIFIED', 'ON_QA', 'VERIFIED',
+                                                    'RELEASE_PENDING', 'CLOSED']
+                            upstream_POST_minus = ['NEW', 'ASSIGNED', 'POST']
+                            if bug.status in downstream_POST_plus and \
+                                            issue.status.name in upstream_POST_minus:
+                                msg = 'The downstream bug %s is at POST+ but the upstream bug %s ' \
+                                      'at POST-. Something is wrong.' % (bug.id, issue.id)
+                                raise RuntimeError(msg)
```
dkliban: you should probably accumulate these error messages like you do with the linking problems. Then raise an Exception at the end with text from all the problems found along the way.

PR #253, opened 2016-11-21
Timeout pulp-smash-runner after 120 minutes

seandst:
These normally take about an hour, unless deadlocking occurs, then
they last for as long as nodepool keeps the slave alive (8 hours),
at which point they're rudely terminated. Going with "abort" here
means we can at least separate out the noise of timed out jobs from
other failure types.


elyezer:
LGTM

PR #254, opened 2016-11-21
Handles permission denied exceptions from Bugzilla

dkliban:


ci/redmine_bugzilla.py
```diff
@@ -38,6 +38,7 @@
 from bugzilla.rhbugzilla import RHBugzilla
 from redmine import Redmine
 import urllib3.contrib.pyopenssl
+import xmlrpclib
```
bmbouter: This dep probably also needs to be installed [here](https://github.com/pulp/pulp_packaging/blob/master/ci/jobs/redmine.yaml#L40) right?

bmbouter:
Can some logging be added before having it `continue` after catching the code 102?

ci/redmine_bugzilla.py
dkliban: no ... it gets installed with the bugzilla libraries

bmbouter: You're right, I just tested this myself

PR #260, opened 2016-12-06
Fixes BZ transition to POST for BZs with multiple Redmine issues

dkliban:


ci/redmine_bugzilla.py
```diff
@@ -141,6 +141,13 @@ def main():
                         if external_bug['type']['description'] == 'Foreman Issue Tracker':
                             # If the bug has an external foreman issue, don't transition the BZ
                             transition_to_post.append(False)
+                        if external_bug['type']['description'] == 'Pulp Redmine':
+                            if bug.status in ['NEW', 'ASSIGNED']:
+                                if issue.status.name in ['MODIFIED', 'ON_QA', 'VERIFIED',
```
bmbouter: This is checking the "current upstream Pulp bug's state" even though these additional external bugs will be associated with another issue. In other words the `issue` is not referencing the necessary thing. Instead of using `issue.status.name` replace it with `external_bug['ext_status']`.

If you do ^ you'll create a second smaller issue which is that the updating of external_bug['ext_status'] does not occur until later. To remedy this I recommend adding a new loop through the external bugs below this one. In fact that loop could contain the code you've moved here, the foreman check above, and handle the moving to POST once this new loop is introduced. What do you think about this?

PR #261, opened 2016-12-12
Update upgrade jobs trigger

elyezer:
Run the upgrade jobs twice a week on every Monday and Thursday.


bmbouter:
I don't know about the syntax, but I like this change.

PR #264, opened 2016-12-15
Install attr package before running Pulp Smash

elyezer:
Pulp Smash is now making use of `getfattr` command, that said, make sure
the attr package, which provides `getfattr` command, is installed.

For more information check related Pulp Smash PR
https://github.com/PulpQE/pulp-smash/pull/455.


Ichimonji10:
Super straightforward change. Thanks! LGTM.

PR #265, opened 2016-12-16
Add Fedora 25 nodes to nodepool configuration

elyezer:


elyezer:
Testing this change showed that Puppet agent is not available to Fedora 25 [1]. We are good to merge this PR but the nodes will only be available when Puppet RPMs are available to F25.

[1] https://yum.puppetlabs.com/

PR #266, opened 2016-12-17
Problem: sphinx_rtd_theme is broken in PyPI

dkliban:
Solution: Install from source using a commit hash that fixes the problem

closes #2486
https://pulp.plan.io/issues/2486

### DIFF
--- a/ci/jobs/docs.yaml
+++ b/ci/jobs/docs.yaml
@@ -51,7 +51,7 @@
             # create a virtualenv in which to install packages needed to build docs
             virtualenv -p /usr/bin/python3 --system-site-packages ~/docs_ve
             source ~/docs_ve/bin/activate
-            pip3 install sphinx sphinx-rtd-theme pyyaml virtualenv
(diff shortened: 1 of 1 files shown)


bmbouter:
Shouldn't we introduce this on 2.10-dev which is the first builder that was broken?

Also some manual things were done to fix 2.10-dev, what were those? For consistency I think reverting those changes would be good and allowing this PR to fix 2.10+.

What do you think?

dkliban:
The change I am introducing here was already manually introduced into the 'docs-builder-2.10-dev' job definition on Jenkins. I would like to update all the doc builders to use the same version of sphinx_rtd_theme.

The older docs have not been affected by this bug because they don't build nightly.

bmbouter:
With JBB, those changes to the 'docs-builder-2.10-dev' job will be lost on the next push right? If that is reverted and this is rebased against 2.10-dev we wouldn't ever loose the change.

bmbouter:
Disregard my request about rebasing, this repo doesn't version in that way.

PR #268, opened 2016-12-20
Improve pre and post upgrade scripts

elyezer:
The improvements are:

* Sync Python content
* Use queries to sync fewer Puppet modules instead of entire Puppet
  Forge
* Use pulp-fixtures repositories instead of demo_repos
* Sync centos5 from an official mirror URL
* Remove repo_1 repository
* Fix file-feed repo feed path
* Add more Docker images to sync


elyezer:
Rebased, will merge as soon as travis is green.

PR #271, opened 2017-01-03
Update prepare_node.sh

seandst:
After a recent change that stopped running our puppet modules, these packages
needed to be manually installed to get the redmine-bugzilla-automation job working.

Since it was a change the prepare_node.sh script that most likely caused that problem,
I think it's probably a good idea to put the fix in the same file.


elyezer:
Going to merge this and rebuild the images

PR #274, opened 2017-01-05
Update pr-tester jobs to install deps previous installed by puppet

seandst:


seandst:
Some notes for...posterity, I guess?

- I've tested this with the main pulp PR tester. I assume it will work with the plugin testers based on their working in a very similar way, but it's untested there.
- As mentioned in the comment in the code, this is based on stuff that puppet used to do. Rather than make a new node type, it seemed better to put all the necessary deps right in the job, and in general apply "band-aids" to these pulp 2 jobs to keep them working rather than make broad changes to how we do our testing in a version approaching its end of life.
- We've already got an ansible playbook for this in pulp 3, so I think we can look forward to opportunities to reuse the devel playbooks in jenkins jobs, in nodepool, or potentially both where it makes sense.

PR #276, opened 2017-01-06
Install attr package on the Pulp server machine

elyezer:
The attr package which provides the `getfattr` executable should be
installed on the Pulp server machine and not on Pulp Smash machine since
the command will be run by Smash on the Pulp machine via SSH.

This will fix 3 failures on automation.


Ichimonji10:
LGTM

Ichimonji10:
For reference, see: https://github.com/PulpQE/pulp-smash/pull/455#issuecomment-267078279

PR #279, opened 2017-01-10
Install attr package for upgrade jobs

elyezer:
The attr package is needed by Pulp Smash in order to run some tests,
make sure the Pulp Server machines for the upgrade jobs have that
package installed.


Ichimonji10:
LGTM

This should fix failures in our upgrade test suite.

PR #280, opened 2017-01-10
Clean up mac addrs in eth0 ifcfg when preparing a node

seandst:


seandst:
Maybe I should have mentioned this in the commit, but this gets the fedora 25 images working for us. The changes shouldn't have any negative impact on other images.

elyezer:
Thank you very much for fixing this.

PR #283, opened 2017-01-18
Use f25 nodes, don't use f23 nodes, remove vanilla labels

seandst:
With the removal of the puppet stuff in nodepool's prepare_node script, the vanilla nodes are now identical to the non-vanilla nodes. We fixed up the PR tester jobs to install the missing stuff that the puppet modules used to be installing, so now the former differences between the vanilla and non-vanilla nodes is captures in the PR tester jobs themselves. So, this commit merges two  things:

1. Replace all f23 jobs with f25 jobs per our fedora support policy
2. Replace all vanilla nodes with non-vanilla nodes and clean out references to vanilla nodes from nodepool's configs

### DIFF
--- a/ci/jobs/pulp-installer.yaml
+++ b/ci/jobs/pulp-installer.yaml
@@ -2,7 +2,7 @@
 # machines for running Pulp Smash.
 - job:
     name: 'pulp-installer'
-    node: 'f25-vanilla-np'
+    node: 'fedora-np'
     description: |
         <p>Job that installs Pulp on the machine identified by the job
         parameter.</p>
--- a/ci/nodepool/nodepool.yaml
+++ b/ci/nodepool/nodepool.yaml
@@ -14,26 +14,25 @@ cron:
   cleanup: '*/1 * * * *'
 
 labels:
+  # unversioned labels used to abstract the distribution version from the job,
+  # useful where the job just wants the latest available fedora, not a specific version.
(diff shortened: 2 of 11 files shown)


seandst:
@elyezer, you'd mentioned to me that there was a job that was using f25-np that could be switched to the unversioned fedora-np, but I don't recall which job it was or if I've included that change in this PR. Do you remember which one it was?

ci/jobs/pulp-installer.yaml
```diff
@@ -2,7 +2,7 @@
 # machines for running Pulp Smash.
 - job:
     name: 'pulp-installer'
-    node: 'f25-vanilla-np'
+    node: 'f25-np'
```
elyezer: This can use the fedora-np

ci/nodepool/nodepool.yaml
```diff
@@ -14,26 +14,31 @@ cron:
...
+  # unversioned docker image, similar to fedora-np, but with docker installed
+  - name: docker-np
+    image: docker-np
     min-ready: 0
     providers:
       - name: cios
     ready-script: fix_hostname.sh
-  - name: f23-docker-np
-    image: f23-docker-np
+  - name: fedora-np
```
elyezer: Missed this on the previous review, but it seems to be duplicated. There is a fedora-np above already.

seandst:
@elyezer changes address, thanks!

elyezer:
@seandst thank you!

PR #285, opened 2017-01-20
2.12 Upgrade Jobs

preethit:
Adding 2.12 to upgrade permutations

### DIFF
--- a/ci/jobs/projects.yaml
+++ b/ci/jobs/projects.yaml
@@ -92,6 +92,8 @@
         - 2.10-nightly
         - 2.11-beta
         - 2.11-nightly
+        - 2.12-beta
+        - 2.12-nightly
     exclude:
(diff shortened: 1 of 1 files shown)


ci/jobs/projects.yaml
```diff
@@ -92,6 +92,8 @@
         - 2.10-nightly
         - 2.11-beta
         - 2.11-nightly
+	- 2.12-beta
```
dkliban: I think this is marked as red because a Tab character is used instead of spaces. I think your editor is probably set to using tabs instead of spaces. Could you please change this to spaces so it's consistent with the rest of the file? Thanks!

elyezer:
Merged as 0d6e93767b6425ef57af1fe32cc8222a2960e4c3

PR #286, opened 2017-01-25
Modernize privilege escalation procedure

Ichimonji10:
Use `become` and `become_method` instead of `sudo` in playbooks. From
the Ansible docs:

> As of 1.9 *become* supersedes the old sudo/su, while still being
> backwards compatible.

See: http://docs.ansible.com/ansible/become.html

### DIFF
--- a/ci/ansible/pulp_coverage.yaml
+++ b/ci/ansible/pulp_coverage.yaml
@@ -2,4 +2,5 @@
 - hosts: all
   roles:
     - role: pulp-coverage
-  sudo: true
+  become: true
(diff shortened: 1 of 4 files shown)


Ichimonji10:
Every time I use Ansible to install Pulp, I see a deprecation warning about the use of `sudo`.

elyezer:
LGTM

elyezer:
Merged as 2ddb173fa17ad763399fab5b9441aa2df90d35ef. Thank you very much @Ichimonji10

PR #287, opened 2017-01-26
Use py.test to run Pulp Smash tests

elyezer:
Update pulp-smash-runner job to use py.test to run the tests. Also
update the generated XML jUnit report file to something more generic so
we can change the tool later and use the same file name.


Ichimonji10:
LGTM.

PR #289, opened 2017-02-02
Pull the current pulp version from setup.py in pr tests

seandst:
We currently pull this from a file that tito drops into the source repo,
which confuses the PR tester on master because this file only gets updated
by tito when we do a release build. The spec file and setup.pys do get
updated regularly, though. Since the spec files are moving to the packaging
repo, it makes sense to ask the setup.py file what the current version is
and use that, rather than depend on a file made by tito after a release.

### DIFF
--- a/ci/jobs/unittests.yaml
+++ b/ci/jobs/unittests.yaml
@@ -69,8 +69,8 @@
 
             set -x
 
-            # get the base pulp version
-            PULP_VERSION=$(grep -Po "(^[\d]+\.[\d]+)" pulp/rel-eng/packages/pulp)
+            # get the base pulp version from the server setup.py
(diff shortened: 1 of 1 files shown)


seandst:
I haven't tested this solution yet in jenkins, but can't come up with a scenario where the setup.py call will fail when used this way.

seandst:
Tested on a master PR:
```
00:01:08.766 [f25-np] $ /bin/bash /tmp/hudson6381978774342966282.sh
00:01:08.770 Setting up after source control extract
00:01:08.791 ++ python2 pulp/server/setup.py --version
00:01:08.792 ++ grep -Po '([\d]+\.[\d]+)'
00:01:09.028 + PULP_VERSION=2.13
00:01:09.029 + echo 'PULP BASE VERSION: 2.13'
00:01:09.029 PULP BASE VERSION: 2.13
```

PR #291, opened 2017-02-06
"Librarize" common git cloning code across scripts

seandst:
Both the merge forward script and build-all have had duplicated
code for a while. The introduction of docs-builder came with a
third place where git cloning was done. I put in a hacky fix in
centralizing the git cloning and update code.

This is that follow-up. :)

The diff is, I think, a case of "It looks worse than it is." The
basic plan I followed was to functionalize behaviors, and then move
any function that isn't "parse_args", "main", etc. up into the lib
modules to be shared by the multiple scripts that want to use them.

To test, build-all, docs-builder, and the update-version-and-merge-forward
scripts should all work as expected, and definitely not explode.

### DIFF
--- a/ci/lib/promote.py
+++ b/ci/lib/promote.py
@@ -1,6 +1,19 @@
+import glob
 import re
+import os
+import shutil
 import subprocess
 import sys
+
+try:
+    # py2
+    from StringIO import StringIO
+except ImportError:
+    # py3
+    from io import StringIO
+
+VERSION_REGEX = "(\s*)(version)(\s*)(=)(\s*)(['\"])(.*)(['\"])(.*)"
+RELEASE_REGEX = "(\s*)(release)(\s*)(=)(\s*)(['\"])(.*)(['\"])(.*)"
 
 
 def get_promotion_chain(git_directory, git_branch, upstream_name='origin', parent_branch=None):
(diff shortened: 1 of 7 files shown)


ci/lib/promote.py
```diff
@@ -1,7 +1,20 @@
...
+import shutil
 import subprocess
 import sys
 
+try:
+    # py2
+    from StringIO import StringIO
+except ImportError:
+    # py3
+    from io import StringIO
```
elyezer: Python 3 ftw

ci/lib/promote.py
```diff
@@ -234,3 +247,215 @@ def merge_forward(git_directory, push=False, parent_branch=None):
...
+        match = re.match(release_regex, line)
+        if match:
+            line = "".join((match.group(1), release, "%{?dist}\n"))
+
+        out_f.write(line)
+
+    in_f.close()
+    out_f.close()
+    shutil.move(spec_file + ".new", spec_file)
+    print("updated %s to %s-%s" % (spec_file, version, release))
```
elyezer: Since you're using print as a function I think it would be better to `from __future__ import print_function`

seandst: You're right, but that change wasn't made in this PR so I'm going to disregard it for now. I think I'd rather just switch these scripts to work under python3, because...they already do :)

The only thing potentially blocking that is whether we run these scripts on rhel7 or not, so I'll make a note to look into this and create/fix any outstanding issues. I think that changing shebangs to python3 and ensuring we build on the fedora-np image and not, say, rhel7, is all it will take.

PR #295, opened 2017-02-13
Rewrite task to use iptables module

Ichimonji10:
Rewrite a task to use the "iptables" Ansible module instead of the
"shell" ansible module. This ensures that changes are only made when
necessary.

### DIFF
--- a/ci/ansible/roles/pulp/tasks/main.yaml
+++ b/ci/ansible/roles/pulp/tasks/main.yaml
@@ -5,7 +5,13 @@
     that: ansible_os_family == "RedHat"
 
 - name: Open firewall ports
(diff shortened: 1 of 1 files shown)


Ichimonji10:
`ansible-playbook -vv` shows that the following iptables rule is added:

    -p tcp -j ACCEPT --destination-port 80 -m state --state NEW

`iptables-save` prints the following rules, among others:

```
-A INPUT -p tcp -m tcp --dport 5671 -m state --state NEW -j ACCEPT
-A INPUT -p tcp -m tcp --dport 5672 -m state --state NEW -j ACCEPT
-A INPUT -p tcp -m tcp --dport 443 -m state --state NEW -j ACCEPT
-A INPUT -p tcp -m tcp --dport 80 -m state --state NEW -j ACCEPT
```

PR #296, opened 2017-02-14
Update path to python-qpid{,-common}

Ichimonji10:
URLs to python-qpid{,-common} packages are hard-coded into the Ansible
configuration file. The given URLs are broken - they point to
non-existent files, and servers correctly return HTTP 404 responses when
asked for those files. Update the URLs.


Ichimonji10:
I have absolutely no understanding of Koji, so if I'm doing something stupid, say so. This change is, however, necessary for me to install Pulp on RHEL 7 systems.

elyezer:
I'm sorry I missed seeing this before submitting #299, those package are not needed anymore, that said they were dropped.

PR #297, opened 2017-02-14
Use Ansible's redhat_subscribe module

Ichimonji10:
The `redhat_subscription` module is expressly designed to manage a
system's subscriptions via `subscription-manager` in an idempotent
manner. Use it instead of the `shell` module.

The current subscription-manager shell command lets one select a pool
via its ID. In contrast, the `redhat_subscription` module lets one
select a pool via its name. This is a **breaking change** for
pulp_packaging users.

### DIFF
--- a/ci/ansible/roles/pulp/tasks/main.yaml
+++ b/ci/ansible/roles/pulp/tasks/main.yaml
@@ -19,25 +19,43 @@
     - 5671
   when: pulp_install_prerequisites
 
-- name: Subscription Manager register
-  shell: "subscription-manager register --force --user={{ rhn_username }} --password={{ rhn_password }}"
+# De-registering and then registering is equivalent to using the
+# `force_register` argument, which was added in Ansible 2.2. We use this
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Is there any reason `--force` is used when registering and subscribing?

elyezer:
> Is there any reason --force is used when registering and subscribing?

That is needed to make sure it will work even if the machine has a subscription already

Ichimonji10:
As per http://docs.ansible.com/ansible/redhat_subscription_module.html, the `force_register` parameter was added in Ansible 2.2. That's not good for us, as we run Ansible from RHEL 6 systems. I can work around this by either removing `force_register` or prepending a step which de-registers the system.

elyezer:
Second option looks promising to me.

Ichimonji10:
Will do!

Ichimonji10:
Verifying...

Ichimonji10:
The updated PR works.

PR #300, opened 2017-02-15
Save IPv4 iptables firewall if modified

Ichimonji10:
This change ensures that changes to the firewall persist across reboots.

### DIFF
--- a/ci/ansible/roles/pulp/handlers/main.yaml
+++ b/ci/ansible/roles/pulp/handlers/main.yaml
@@ -24,3 +24,6 @@
   service:
     name: goferd
     state: restarted
+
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Only RHEL 6 uses iptables. RHEL 7, Fedora 24 and Fedora 25 use firewalld. Thus, this change is only really effective for RHEL 6.

I think calls to iptables on RHEL 7, Fedora 24 and Fedora 25 are intercepted by firewalld, or something like that. Regardless of the exact reason, we should update *all* of the firewall handling code so that iptables is used on RHEL 6 and firewalld is used for the other distros.

elyezer:
:+1:

PR #302, opened 2017-02-15
Problem: incorrect BZ status when FailedQA

dkliban:
Solution: check for the FailedQA flag when looking at the status of
BZs. If FailedQA is set, add a needinfo flag with the requestee set
to the email address of the person assigned in Redmine issue.

### DIFF
--- a/ci/redmine_bugzilla.py
+++ b/ci/redmine_bugzilla.py
@@ -120,6 +120,8 @@ def main():
     ext_bug_record = ''
     downstream_state_issue_record = ''
     downstream_changes = ''
+    new_failed_qa_record = ''
+    failed_qa_bugzillas = []
 
     for issue in redmine_issues:
         for custom_field in issue.custom_fields.resources:
@@ -128,6 +130,8 @@ def main():
                     continue
                 for bug_id in [int(id_str) for id_str in custom_field['value'].split(',')]:
                     links_back = False
+                    if bug_id in failed_qa_bugzillas:
+                        continue
                     try:
                         bug = BZ.getbug(bug_id)
                     except xmlrpclib.Fault as e:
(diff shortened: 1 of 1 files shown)


ci/redmine_bugzilla.py
```diff
@@ -181,7 +185,28 @@ def main():
...
+                                        for flag in bug.flags:
+                                            if flag['name'] == 'needinfo' and flag['requestee'] == needinfo_email:
+                                                needinfo = False
+                                        if needinfo:
+                                            BZ.update_flags(bug.id, [{ "name": "needinfo",
+                                                                       "status": "?",
+                                                                       "requestee": needinfo_email,
+                                                                       "new": True}])
+                                            bug.addcomment("Requesting needsinfo from upstream " \
+                                                           "developer %s because 'FailQA' flag is " \
```
bmbouter: s/because/because the/

ci/redmine_bugzilla.py
```diff
@@ -181,7 +185,28 @@ def main():
...
+                                                           "set." % needinfo_email)
+                                            new_failed_qa_record += "%s\n" % msg
+                                        print msg
```
bmbouter: Is this supposed to be here? The report is nicely printed below on L#267

ci/redmine_bugzilla.py
```diff
@@ -181,7 +185,28 @@ def main():
                             if bug.status in ['NEW', 'ASSIGNED']:
                                 if external_bug['ext_status'] in ['MODIFIED', 'ON_QA', 'VERIFIED',
                                                          'CLOSED - CURRENTRELEASE']:
-                                    transition_to_post.append(True)
+                                    if 'FailedQA' in bug.cf_verified:
```
bmbouter: Do we want this FailedQA branch guarded by L# 185 and L#186? I think maybe we don't, but I'm not totally sure. If we don't it would be moved above L#185 and the old cold L#181-184 would be part of the else block.

dkliban: I expect that when an issue has FailedQA flag, it is in the ASSIGNED state.

PR #303, opened 2017-02-15
Change rhn_pool to rhn_poolid

werwty:
ansible is passing in rhn_poolid instead of rhn_pool


elyezer:
This was right, the ansible module expects the `pool` parameter which is the pool name not the pool id

PR #304, opened 2017-02-16
Use RHN Pool name instead of Pool ID

elyezer:
Revert the changes so the ansible playbook uses the rhn_playbook
variable and update the job definition to proper pass that renamed
parameter.

### DIFF
--- a/ci/ansible/roles/pulp/tasks/main.yaml
+++ b/ci/ansible/roles/pulp/tasks/main.yaml
@@ -29,7 +29,7 @@
     state: absent
   when: >
     ansible_distribution == "RedHat" and
-    rhn_poolid is defined and
+    rhn_pool is defined and
     (rhn_username is defined and rhn_password is defined) or
     (rhn_activation_key is defined and rhn_organization is defined)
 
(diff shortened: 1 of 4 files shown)


elyezer:
With these changes `git grep -i rhn_poolid` don't return anything

ci/ansible/roles/pulp/tasks/main.yaml
```diff
@@ -29,18 +29,18 @@
     state: absent
   when: >
     ansible_distribution == "RedHat" and
-    rhn_poolid is defined and
+    rhn_pool is defined and
```
Ichimonji10: Ugh. I don't know how I missed this.

Ichimonji10:
Merging.

Ichimonji10:
Just to make sure: after applying this change, we'll pass in a regex like `'^SKU Name$'` when Jenkins provisions machines, right?

elyezer:
> Just to make sure: after applying this change, we'll pass in a regex like '^SKU Name$' when Jenkins provisions machines, right?

Yes, the recent runs already had that in place and worked fine. Thank you for making sure everything is alright.

PR #310, opened 2017-02-22
Problem: bz/redmine jenkins job owners not accurate

dkliban:
Solution: create a new ownership group for bz/redmine job

### DIFF
--- a/ci/jobs/macros.yaml
+++ b/ci/jobs/macros.yaml
@@ -67,6 +67,16 @@
                 - semyers
                 - ttereshc
 
+# Ownership that defines owners for Bugzilla/Redmine integration job
+- property:
+    name: bz-redmine-ownership
+    properties:
+        - ownership:
+            owner: dadavis
+            co-owners:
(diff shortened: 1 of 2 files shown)


ci/jobs/macros.yaml
```diff
@@ -67,6 +67,16 @@
                 - semyers
                 - ttereshc
 
+# Ownership that defines owners for Bugzilla/Redmine integration job
+- property:
+    name: bz-redmine-ownership
+    properties:
+        - ownership:
+            owner: dadavis
```
bmbouter: fyi, there are two accounts in Jenkins for @daviddavis There is dadavis and daviddavis. I'm not sure which is prefered. Maybe remove the one that isn't used?

daviddavis: I don't know how 'daviddavis' got created or how to log into it. My kerberos id though is 'dadavis' so I'd be in favor of removing 'daviddavis'.

PR #317, opened 2017-02-27
Use firewalld instead of iptables on most systems

Ichimonji10:
The primary IPv4 firewall configuration tool on RHEL 6 is iptables. The
primary IPv4 firewall configuration tool on RHEL 7, Fedora 24 and Fedora
25 is firewalld. Update the Ansible roles for installing Pulp so that
firewalld is used in most cases, and add "python-firewalld" as a
dependency. The upshot is that firewall rules fore these distributions
are now persistent across system and service restarts.

### DIFF
--- a/ci/ansible/roles/pulp/tasks/main.yaml
+++ b/ci/ansible/roles/pulp/tasks/main.yaml
@@ -4,7 +4,7 @@
   assert:
     that: ansible_os_family == "RedHat"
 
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Tested with Fedora 24, Fedora 25, RHEL 6.8 and RHEL 7.3. The firewall configuration persists across reboots for all systems.

PR #318, opened 2017-03-02
Install firewalld before trying to update firewall rules

seandst:
Recent builds on fedora are failing because the iptables step attempts
to use firewalld's python bindings to update the firewall and fails
because firewalld is missing.

### DIFF
--- a/ci/jobs/scripts/pulp-install.sh
+++ b/ci/jobs/scripts/pulp-install.sh
@@ -1,4 +1,4 @@
-sudo yum -y install ansible attr git libselinux-python python-firewall
+sudo yum -y install ansible attr git libselinux-python firewalld python-firewall
 echo 'localhost' > hosts
 source "${RHN_CREDENTIALS}"
 ansible-playbook --connection local -i hosts ci/ansible/pulp_server.yaml \


Ichimonji10:
That's so weird. I thought fedora included firewalld by default. I'm not sure how I missed that. Sorry, all.

seandst:
This looks incomplete. Even after installing firewalld, it looks like the service needs to be enabled and started before the ansible task will work. Furthermore, I didn't check to see if firewalld is even installable in el6, so those nightlies will likely fail around the same spot.

seandst:
For now, I think all the dev jobs are broken as a result of this and https://github.com/pulp/pulp_packaging/commit/a9ef8a03dc60ee48d3cda4375b566cd88720360a

Ichimonji10:
firewalld isn't available on el6. That should be fine, though - yum returns a status code of zero as long as *any* of the packages it's been asked to install can be installed.

Furthermore, on Fedora 24 and 25 server edition, firewalld is installed and enabled by default.

Ichimonji10:
After doing some digging around, I can definitely say that Jenkins is not executing this updated script. Take a look at the console logs from any Jenkins run, and you'll see what I mean. Or, take a look at the configuration for any failing Jenkins job, and you'll see that the script that Jenkins is configured to run differs from this script.

Does JJB need to be executed in some way?

PR #320, opened 2017-03-10
Refactor firewall rule declarations

Ichimonji10:
Do the following:

* Use service names instead of port numbers where possible.
* Pass port/transport pairs to the firewall management tasks. This makes
  it easier to add firewall rules that allow non-tcp connections.


Ichimonji10:
I'm trying to get rid of a DHCP lease expiration issue, and this was one of the steps I took in creating a possible solution. I've tested this commit all by itself, and it works fine.

elyezer:
Rebased and merged as 6e4de4a5311e08586f8c83b6c9f39594360210ad

PR #321, opened 2017-03-14
Problem: script fails to set needinfo flag on bugzillas

dkliban:
Solution: update script to use the 2.0 python-bugzilla API for setting
needinfo flag.

### DIFF
--- a/ci/redmine_bugzilla.py
+++ b/ci/redmine_bugzilla.py
@@ -200,10 +200,12 @@ def main():
                                                             flag['requestee'] == needinfo_email:
                                                 needinfo = False
                                         if needinfo:
-                                            BZ.update_flags(bug.id, [{ "name": "needinfo",
-                                                                       "status": "?",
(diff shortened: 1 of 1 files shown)


ci/redmine_bugzilla.py
```diff
@@ -200,10 +200,12 @@ def main():
...
-                                            BZ.update_flags(bug.id, [{ "name": "needinfo",
-                                                                       "status": "?",
-                                                                       "requestee": needinfo_email,
-                                                                       "new": True}])
+                                            flag = { "name": "needinfo",
+                                                     "status": "?",
+                                                     "requestee": needinfo_email,
+                                                     "new": True}
+                                            updates = BZ.build_update(status=bug.status, flags = [flag])
+                                            BZ.update_bugs(bug.id, updates)
```
seandst: So, just some fun API changes, it looks like?

dkliban: yep

PR #325, opened 2017-03-20
Updates for el7 jenkies

seandst:
- irc notification disabled
- obsolete credentials removed
- explicit shebangs removed
- overly restrictive bash settings removed from ssh-setup script

All jobs should be updated after merge.

### DIFF
--- a/ci/jobs/pulp-dev.yaml
+++ b/ci/jobs/pulp-dev.yaml
@@ -23,6 +23,8 @@
         - shell:
             !include-raw-escape:
                 - 'ssh-setup.sh'
+        - shell:
+            !include-raw-escape:
                 - 'pulp-install.sh'
                 - 'pulp-coverage-setup.sh'
                 - 'pulp-smash-parameters.sh'
@@ -96,7 +98,6 @@
             artifacts: "*.tar.gz"
             allow-empty: true
         - email-notify-owners
-        - irc-notify-all-summary
         - mark-node-offline
 
 - job:
--- a/ci/jobs/build-automation.yaml
+++ b/ci/jobs/build-automation.yaml
@@ -75,7 +75,6 @@
           # do not need to ensure that there is a promote job for every release config
           projects:
             - "build-automation-promote-{release_config}"
-      - irc-notify-all-summary
       - email-notify-owners
       - mark-node-offline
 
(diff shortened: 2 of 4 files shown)


ci/jobs/scripts/post-upgrade.sh
```diff
@@ -1,4 +1,3 @@
-#!/bin/bash
```
elyezer: Please leave the shebangs, this scrips can be used standalone.

ci/jobs/scripts/pre-upgrade.sh
```diff
@@ -1,4 +1,3 @@
-#!/bin/bash
```
elyezer: Please leave the shebangs, this scrips can be used standalone.

ci/jobs/scripts/ssh-setup.sh
```diff
@@ -1,5 +1,2 @@
-#!/usr/bin/env bash
-set -euo pipefail
```
elyezer: This will ensure the script does not fail, please leave it.

ci/jobs/scripts/post-upgrade.sh
seandst: It isn't marked executable, so no it can't.

ci/jobs/scripts/pre-upgrade.sh
seandst: It isn't marked executable, so no it can't.

ci/jobs/scripts/ssh-setup.sh
seandst: In jobs, this script is included in shell jobs with other scripts, resulting in unexpected behavior when the other scripts run. At the very least, it should be run in isolation.

seandst: Also this was required during some testing on the el7 master, and its removal did not cause the script to fail. What exactly about these settings ensures the script does not fail?

seandst:
pr updated based on comments, we can tackle the shell issues later. They're all going to get addressed as we continue to lean on ansible for our ci automation.

ci/jobs/scripts/ssh-setup.sh
Ichimonji10: If any command produces a non-zero return code, even in the middle of a piped sequence, then this will cause bash to abort execution and return that code. Also, if an uninitialized variable is used, then bash will abort execution. This basically lets us abide by the idea that "errors should never pass silently unless explicitly silenced."

ci/jobs/pulp-dev.yaml
```diff
@@ -23,6 +23,8 @@
         - shell:
             !include-raw-escape:
                 - 'ssh-setup.sh'
+        - shell:
```
elyezer: We could have everything on the same shell build step, but it is fine to leave this way.

seandst:
Jenkins jobs have been updated.

PR #326, opened 2017-03-27
Jenkins Job Tweaks

seandst:
- include pulpadmin credential where jenkins-ssh-credentials is used
  - jobs that are using jenkins-ssh-credentials expect to have access to
    the fedorapeople site. These jobs are only working because we added
    that key to the jenkins user on the build box, this explicitly adds
    those credentials to the job, allowing it to run anywhere.
- use the clean workspace wrapper instead of 'rm -rf *' in the sync triage
  job
- add timestamps to job build logs by default

### DIFF
--- a/ci/jobs/build-automation.yaml
+++ b/ci/jobs/build-automation.yaml
@@ -35,6 +35,7 @@
             timeout: 240
             timeout-var: 'BUILD_TIMEOUT'
             fail: true
+        - timestamps
     builders:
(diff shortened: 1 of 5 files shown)


seandst:
Just running these through jenkins and making sure things are working as expected before I merge.

seandst:
> Just running these through jenkins and making sure things are working as expected before I merge.

...and they do! :D

PR #330, opened 2017-04-07
Move subscription-manager tasks into separate role

Ichimonji10:
One of the roles called by the "pulp_server" playbook is the "pulp"
role. This role's job is to install and configure Pulp.  This role has
two sets of tasks: `tasks/main.yaml` and `tasks/pulp_server.yaml`.  The
former is almost entirely concerned with configuring
subscription-manager, and the latter does most of the actual work of
installing and configuring Pulp.

This seems a little strange: shouldn't the "pulp" role's `main.yaml`
install and configure Pulp? It's also a little problematic: there's no
easy way to skip the subscription-manager configuration, and this is
reflected by the large number of `when: ansible_distribution ==
'RedHat'` expressions in the `main.yaml` file.

Do the following:

* Make a new role named `subscription-manager`, and move the
  subscription-manager management logic into it.
* Move the firewall management code from the subscription-manager tasks
  to the pulp management tasks. After all, firewall management and
  configuration is needed to successfully get Pulp running, regardless
  of whether subscription-manager is present.
* Move the `pulp_install_prerequisites` variable. It controls whether
  the target system's firewall will be configured, and it seems like
  something that should always be done.
* Drop the `pulp_install_server` variable. If the Pulp server should be
  installed, that can now be accomplished by... running the `pulp` role.


Ichimonji10:
Tested with F24, F25 and RHEL 7 with `ansible-playbook pulp_server.yaml -i hosts`, where `hosts` is the following:

```
fedora-24-pulp-2-12 pulp_version='"2.12"'
fedora-25-pulp-2-12 pulp_version='"2.12"'
rhel-7-3-pulp-2-12 pulp_version='"2.12"'
```

I'm also currently testing against RHEL 6, with the following patch:

```diff
diff --git a/ci/ansible/roles/pulp/tasks/main.yaml b/ci/ansible/roles/pulp/tasks/main.yaml
index c055a43..ba495b9 100644
--- a/ci/ansible/roles/pulp/tasks/main.yaml
+++ b/ci/ansible/roles/pulp/tasks/main.yaml
@@ -57,6 +57,8 @@
 - name: Start and enable MongoDB server service
   service: name=mongod state=started enabled=yes

+- meta: end_play
+
 - name: Setup qpid custom repo
   get_url: url=https://copr.fedorainfracloud.org/coprs/g/qpid/qpid/repo/epel-6/irina-qpid-epel-6.repo dest=/etc/yum.repos.d/copr-qpid.repo
   when:
```

Ichimonji10:
Tested with F24, F25 and RHEL 7 with `ansible-playbook pulp_server.yaml -i hosts`, where `hosts` is the following:

```
fedora-24-pulp-2-12 pulp_version='"2.12"'
fedora-25-pulp-2-12 pulp_version='"2.12"'
rhel-7-3-pulp-2-12 pulp_version='"2.12"'
```

I've also partially tested against RHEL 6, with the following patch:

```
diff --git a/ci/ansible/roles/pulp/tasks/main.yaml b/ci/ansible/roles/pulp/tasks/main.yaml
index c055a43..ba495b9 100644
--- a/ci/ansible/roles/pulp/tasks/main.yaml
+++ b/ci/ansible/roles/pulp/tasks/main.yaml
@@ -57,6 +57,8 @@
 - name: Start and enable MongoDB server service
   service: name=mongod state=started enabled=yes

+- meta: end_play
+
 - name: Setup qpid custom repo
   get_url: url=https://copr.fedorainfracloud.org/coprs/g/qpid/qpid/repo/epel-6/irina-qpid-epel-6.repo dest=/etc/yum.repos.d/copr-qpid.repo
   when:
```

PR #331, opened 2017-04-10
Adds a Rebase Analysis nightly job

bmbouter:
This automation is a Jenkins Job and a Python script that determines
what downstream bugs will be disincluded if downstream rebases to a
given upstream version.

### DIFF
--- a/ci/jobs/rebase_analysis.yaml
+++ b/ci/jobs/rebase_analysis.yaml
@@ -1,8 +1,8 @@
-# This Jenkins job runs a script which performs automation related to Redmine
-# and Bugzilla. See https://github.com/pulp/pulp_packaging/ci/redmine_bugzilla.py
+# This Jenkins job runs a script which analyzes the gap between upstream and downstream by querying
+# Redmine and Bugzilla. See https://github.com/pulp/pulp_packaging/ci/redmine_bugzilla.py
 
 - job:
-    name: 'redmine-bugzilla-automation'
+    name: 'rebase-analysis'
     defaults: ci-workflow-runtest
     node: 'f25-np'
     properties:
@@ -16,7 +16,7 @@
             skip-tag: true
             wipe-workspace: false
     triggers:
-        - timed: "0,30 * * * *"
+        - timed: "@midnight"
     wrappers:
         - credentials-binding:
             - file:
@@ -24,7 +24,7 @@
                 variable: REDMINE_BUGZILLA_CONF
         - timeout:
             # Timeout in minutes
-            timeout: 30
+            timeout: 60
             timeout-var: 'BUILD_TIMEOUT'
             fail: true
     builders:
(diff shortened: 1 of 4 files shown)


ci/jobs/rebase_analysis.yaml
```diff
@@ -0,0 +1,51 @@
...
+            set -x
+            env
+
+            cd $WORKSPACE/
+
+            sudo pip install python-bugzilla python-redmine certifi
+
+            # Here due to InsecurePlatformWarning
+            # https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning
+            #sudo pip install ndg-httpsclient pyasn1 pyopenssl urllib3
```
elyezer: Should this line be commented?

ci/jobs/rebase_analysis.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+            python pulp_packaging/ci/rebase_analysis.py
+
+    publishers:
+      # Take the node offline so that another build doesn't pile on
```
elyezer: I think this comment should be before the `- mark-node-offline` line. Emailing owners is not related to what it states.

ci/rebase_analysis.py
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env python
```
elyezer: Since `fedora-np` node is being used, maybe it would be better to say python2 here. That will ensure Python 2 and will be noisy if Python 2 is not available at some point.

ci/jobs/rebase_analysis.yaml
bmbouter: I was leaving this in because I thought I would need it after I pushed the jobs, but I can just bring it back from version control if that happens. I am removing it with next force push.

ci/rebase_analysis.py
bmbouter: I'm updating it to use `#!/usr/bin/env python2` with next force push. Let me know if that isn't right.

ci/jobs/rebase_analysis.yaml
```diff
@@ -0,0 +1,51 @@
+# This Jenkins job runs a script which analyzes the gap between upstream and downstream by querying
+# Redmine and Bugzilla. See https://github.com/pulp/pulp_packaging/ci/redmine_bugzilla.py
+
+- job-template:
```
dkliban: i think this is a job and not a job-template.

ci/jobs/rebase_analysis.yaml
bmbouter: good idea; fixed in next force push. ty

ci/jobs/rebase_analysis.yaml
bmbouter: So `s/-template//`  ?

dkliban: yes

bmbouter: I copied this from the redmin.yaml which provides the redmine-bugzilla-automation job. I updated this new job and that one to be s/-template//.

bmbouter:
@dkliban I pushed all changes can you re-review please?

PR #332, opened 2017-04-10
Consolidate defaults for privilege escalation

Ichimonji10:
Identical privilege escalation options are defined in all of the
`pulp_*` playbooks. Move them into `ansible.cfg`.

### DIFF
--- a/ci/ansible/ansible.cfg
+++ b/ci/ansible/ansible.cfg
@@ -1,2 +1,6 @@
+# See: docs.ansible.com/ansible/intro_configuration.html
 [defaults]
-remote_user=root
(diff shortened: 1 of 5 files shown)


Ichimonji10:
Tested in the same manner as https://github.com/pulp/pulp_packaging/pull/330#issuecomment-292606475.

ci/ansible/ansible.cfg
```diff
@@ -1,2 +1,6 @@
+# See: docs.ansible.com/ansible/intro_configuration.html
```
elyezer: :heart:

PR #339, opened 2017-04-11
Update release configs for 2.13

pcreech:
Update release configs for 2.13 release

### DIFF
--- a/ci/config/releases/2.13-dev.yaml
+++ b/ci/config/releases/2.13-dev.yaml
@@ -1,33 +1,33 @@
-koji-target-prefix: pulp-2.14
-rsync-target-dir: 2.14/stage
-rsync-tested-dir: 2.14/dev
+koji-target-prefix: pulp-2.13
+rsync-target-dir: 2.13/stage
+rsync-tested-dir: 2.13/dev
 repositories:
   - name: pulp
     external_deps: deps/external_deps.json
     git_url: git@github.com:pulp/pulp.git
-    git_branch: master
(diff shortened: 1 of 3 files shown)


ci/config/releases/2.13-dev.yaml
```diff
@@ -0,0 +1,33 @@
+koji-target-prefix: pulp-2.13
+rsync-target-dir: 2.13/stage
+rsync-tested-dir: 2.13/dev
+repositories:
+  - name: pulp
+    external_deps: deps/external_deps.json
+    git_url: git@github.com:pulp/pulp.git
+    git_branch: 2.13-dev
+    version: 2.13.0-0.1.alpha
```
seandst: https://docs.pulpproject.org/en/nightly/dev-guide/contributing/building.html#updating-versions says
```...the versions of packages building from -dev branches in the x.y-dev config should be higher than the versions of those same pages in the corresponding x.y-build and x.y-release configs.```

Since the build version is 2.13.0-0.1.beta, the dev version should be 2.13.1-0.1.alpha. This applies to anything being released here. Since the x.y.0 release is what creates an x.y-dev branch, I think there's no instance where an x.y-dev config would contain an x.y.0 version (The "dev" builds of x.y.0 come from master prior to branching)

pcreech:
Updated dev to be z+1
Also modified pulp-ostree to the version in 2.12.2, as there will be no changes between 2.12.2 and 2.13 for ostree.

PR #347, opened 2017-04-18
Replace deprecated directive

Ichimonji10:
Ansible's "service" module allows one to declare which state a service
should be in. The allowed states are:

* started
* stopped
* restarted
* reloaded

A state of "running" can also be declared. This state is deprecated and
will be removed in Ansible 2.7. Replace instances of "running" with
"started."


Ichimonji10:
I successfully installed Pulp on an F24 system with this new playbook. Also, I spun up a RHEL 6 system and verified that its version of Ansible supports the "started" keyword.

PR #357, opened 2017-05-03
Remove docs jobs for 2.12 branches

goosemania:


elyezer:
Does this means that we are not going to have any more releases on 2.12.z?

goosemania:
@elyezer , because 2.13.0 is released, I will expect only 2.13.z and no 2.12.z releases.

elyezer:
Thanks for the clarification

PR #358, opened 2017-05-08
Let templates handle undefined variables

Ichimonji10:
The pulp-smash role has a template for the Pulp Smash settings file.
This template attempts to deal with undefined variables with the
following Jinja block:

    {% if my_var %}
    use my_var
    {% endif %}

Unfortunately, this doesn't work as intended. If `my_var` doesn't exist,
an exception is raised. Fix this.


Ichimonji10:
:tada:

PR #360, opened 2017-05-09
Refactor the "pulp-smash" role

Ichimonji10:
Refactor `ci/ansible/roles/pulp-smash/tasks/main.yaml`. Do at least the
following:

* Make the role compatible with RHEL 7.
* Ensure Python 3 is used for testing, not Python 2. (Pulp Smash
  requires Python 3.4 or newer.) Drop all dependencies on Python 2.
* Ensure Git is present. It was formerly assumed to be present.
* Use task-specific modules instead of the shell module where possible.
  This improves the idempotency of the pulp-smash role.
* Where the shell module must be used, make the commands as idempotent
  as possible.

### DIFF
--- a/ci/ansible/roles/pulp-smash/tasks/main.yaml
+++ b/ci/ansible/roles/pulp-smash/tasks/main.yaml
@@ -1,30 +1,81 @@
 ---
+# Pulp Smash requires Python 3.4+, and we want to create a Python 3 virtualenv.
+# Unfortunately, there's some problems on RHEL 7. (And some other distros.)
+# First, `virtualenv -p python3 ...` breaks due to a bug in virtualenv 1.10. [1]
+# Second, `python3 -m venv ...` breaks when installing pip. [2]
+#
+# This Ansible playbook works around the issue by creating a Python 3 virtualenv
+# without pip, and then installing pip into the virtualenv with a script
+# provided by PyPA. This solution is chosen because it's low impact. Alternative
+# solutions require doing things like installing an updated version of pip
+# system-wide with pip, instead of the system package manager.
+#
+# [1] https://github.com/pypa/virtualenv/issues/463#issuecomment-41204800
+# [2] http://stackoverflow.com/a/26314477
 
-- name: Remove previous Pulp Smash repository
-  shell: rm -rf pulp-smash
+- block:
 
-- name: Clone Pulp Smash repository
-  git: repo=https://github.com/PulpQE/pulp-smash.git depth=1 dest=pulp-smash
+  - name: Ensure Python 3 is present (non-RHEL)
+    action: "{{ ansible_pkg_mgr }} name=python3 state=present"
 
-- name: Install required OS packages
-  action: "{{ ansible_pkg_mgr }} name={{ item }} state=present"
-  with_items:
-    - python
-    - python-pip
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against a matrix of six systems, where one axis is F24, F25 and RHEL 7, and the other axis is Pulp 2.13 and 2.14.

ci/ansible/roles/pulp-smash/tasks/main.yaml
```diff
@@ -1,30 +1,81 @@
...
-    - python-pip
-    - python-virtualenv
+  - name: Ensure a Pulp Smash virtualenv is available (non-RHEL)
+    command: python3 -m venv "{{ ansible_user_dir }}/.virtualenvs/pulp-smash"
+    args:
+      creates:  "{{ ansible_user_dir }}/.virtualenvs/pulp-smash"
 
-- name: Create Pulp Smash virtual environment
-  shell: virtualenv venv chdir=pulp-smash
+  when: not ansible_distribution == "RedHat"
```
elyezer: What about `ansible_distribution != "RedHat"`

ci/ansible/roles/pulp-smash/tasks/main.yaml
```diff
@@ -1,30 +1,81 @@
...
+- name: Ensure a Pulp Smash configuration file is present
+  template:
+    src: settings.j2
+    dest: "{{ ansible_user_dir }}/.config/pulp_smash/settings.json"
 
 - name: Run Pulp Smash
-  shell: XDG_CONFIG_DIRS=. venv/bin/py.test --junit-xml=report.xml pulp_smash/tests chdir=pulp-smash
+  shell: >
+    source {{ ansible_user_dir }}/.virtualenvs/pulp-smash/bin/activate
+    && python -m pytest --pyargs pulp_smash.tests
```
elyezer: we need the junit report otherwise all jobs will break.

Ichimonji10: Whoops!

ci/ansible/roles/pulp-smash/tasks/main.yaml
Ichimonji10: That makes more sense.

Ichimonji10:
Updated. I've tested the changes, and can confirm that they work.

Note that currently, `report.xml` is placed at `~/report.xml`, not `~/pulp-smash/report.xml`. Some changes may need to be made to ensure that this file is found and collected. I'm looking around to find how to fix this, but if you know, say.

Ichimonji10:
Waaaait a sec. Should I be editing `ci/jobs/pulp-dev.yaml` instead?

Ichimonji10:
Even better question: can we merge this PR and work toward having the pulp-smash-runner job defined in `ci/jobs/pulp-dev.yaml` run this Ansible job? One of the problems with our current setup (IMO) is that it's hard to test changes to our jobs. Moving logic from JJB to Ansible makes changes easier to test.

elyezer:
@Ichimonji10 sure, the idea on having the junit is to make pulp-smash-runner finally able to run this playbook. It was not using it because previously it was not generating the junit report needed by the automation jobs.

PR #362, opened 2017-05-12
Make pulp-smash-runner use pulp-smash role to install and configure Pulp Smash

elyezer:
Added two commits:

1. Do not run the Pulp Smash tests from within the pulp-smash role. That is because Ansible is meant to be used to configure the system and running the tests is something the the user would do many many times and might choose which options are better for each run.
2. Make the pulp-smash-runner use the pulp-smash role to setup and configure Pulp Smash and then run the tests.


elyezer:
Everything is working properly, let's get this merged.

PR #364, opened 2017-05-16
Make Pulp Smash settings template more robust

Ichimonji10:
The "verify" setting in Pulp Smash's `settings.json` file may be either
a boolean or a string path to a certificate. Let the Jinja2 template
which generates this file handle either case.


Ichimonji10:
I've manually tested this change by running:

```sh
ansible-playbook pulp_smash.yaml [...] -e pulp_smash_verify=true
ansible-playbook pulp_smash.yaml [...] -e pulp_smash_verify=/etc/pki/ca-trust/source/anchors/cacert.pem
```

PR #365, opened 2017-05-16
Conditionally install pip into pulp smash venvs

Ichimonji10:
Update the `pulp-smash-virtualenv` role so that the command to install
pip only runs when a new pip-less virtualenv is created. This change
makes the `pulp-smash-virtualenv` role 100% idempotent.


Ichimonji10:
Tested against a matrix of six VMs, where one axis is F24, F25 and RHEL 7, and the other axis is Pulp 2.13 and 2.14 nightly. To test this, I removed `~/.virtualenvs` from each system, then ran `ansible-playbook pulp_smash.yaml ...` twice. The first time, the virtualenvs were fully created with pip. The second time, the virtualenvs were left alone, and no tasks resulted in a "changed" state. I also ran some Pulp Smash tests after each `ansible-playbook` run.

PR #368, opened 2017-05-22
Adds some notes I had in a local text file

bmbouter:


bmbouter:
You can see the rendered version here:  https://github.com/bmbouter/pulp_packaging/blob/add-jjb-notes/ci/jobs/README.md

ci/jobs/README.md
```diff
@@ -75,3 +75,65 @@ Once you have the UUID of interest, add it into your job template. Each job temp
...
+
+```python
+import ssl
+context = ssl._create_unverified_context()
+response = urlopen(req, context=context, timeout=self.timeout).read()
+```
+
+In my case the __init__.py file actually lived at:
+
+```home/vagrant/.virtualenvs/jjb_env/lib/python2.7/site-packages/jenkins/__init__.py```
```
goosemania: Slash at the beginning is missing.

ci/jobs/README.md
```diff
@@ -75,3 +75,65 @@ Once you have the UUID of interest, add it into your job template. Each job temp
...
+Update the jobs by job name. Make sure you are at the top of the repo. For example:
+
+```commandline
+jenkins-jobs --ignore-cache update ci/jobs 'docs-builder-*'
+```
+
+Update all jobs using `'*'`. Make sure you are at the top of the repo. For example:
+
+```commandline
+jenkins-jobs --ignore-cache update ci/jobs '*'
```
elyezer: Passing `'*'` is not required, this has the same effect: `jenkins-jobs --ignore-cache update ci/jobs`

ci/jobs/README.md
```diff
@@ -75,3 +75,65 @@ Once you have the UUID of interest, add it into your job template. Each job temp
...
+
+To workaround SSL errors patch line 431 of ```/usr/lib/python2.7/site-packages/jenkins/__init__.py```
+
+```python
+import ssl
+context = ssl._create_unverified_context()
+response = urlopen(req, context=context, timeout=self.timeout).read()
+```
+
+In my case the __init__.py file actually lived at:
```
goosemania: Double underscore should be escaped with '\\' to be displayed correctly I think

ci/jobs/README.md
bmbouter: ty. updated in next push.

ci/jobs/README.md
Ichimonji10: :+1: The escaping rules can be tricky. Wrapping the filename in double backticks is another solution.

bmbouter: good call, fixed in next push.

ci/jobs/README.md
bmbouter: ty, fixed in next push

ci/jobs/README.md
```diff
@@ -75,3 +75,65 @@ Once you have the UUID of interest, add it into your job template. Each job temp
...
+```
+
+SSL Issues
+----------
+
+If you don't have the CA that secures our Jenkins instance in your system CA pack JJB will complain
+about SSL issues. To fix this add the CA to your system CA pack. Alternatively you can modify the
+JJB check to allow insecure usage. Warning: allow ***insecure at your own risk***.
+
+To workaround SSL errors patch line 431 of ```/usr/lib/python2.7/site-packages/jenkins/__init__.py```
```
Ichimonji10: Minor point: triple backticks indicate a block of code, and sngle backticks indicate inline code.

bmbouter: good point, fixed in next push

ci/jobs/README.md
```diff
@@ -75,3 +75,65 @@ Once you have the UUID of interest, add it into your job template. Each job temp
...
+[job_builder]
+include_path=ci/jobs/scripts
+```
+
+Example Usage
+-------------
+
+Test the potential job changes. Ensure sure you are at the top of the repo.
+
+```commandline
```
Ichimonji10: Minor point: `commandline` isn't recognized as a valid language by github or the `markdown-calibre` parsers, nor by any other parser I know of. I think you mean `sh`.

ci/jobs/README.md
```diff
@@ -75,3 +75,70 @@ Once you have the UUID of interest, add it into your job template. Each job temp
...
+mkdir ~/.config/jenkins_jobs/
+vi ~/.config/jenkins_jobs/jenkins_jobs.ini
+```
+
+Here is an example config:
+```ini
+[jenkins]
+user=bbouters
+password=5cxr6230dptzp6l7oqga03chupnqa8f5
+url=https://pulp-jenkins.rhev-ci-vms.eng.rdu2.redhat.com
```
Ichimonji10: How about using a non-internal URL, like `pulp-jenkins.example.com`?

Ichimonji10: Is this an internal-only URL? If so, would a URL like pulp-jenkins.example.com be better?

bmbouter: It is internal only, but I still think the real info here is the best. We use the links in irc often so I don't think it's a meaningful secret. I think the actual issue is that our Jenkins isn't public and it's useful to include this info here for devs who do have access.

ci/jobs/README.md
bmbouter: switching it to `sh` in next push, thanks.

ci/jobs/README.md
```diff
@@ -75,3 +75,70 @@ Once you have the UUID of interest, add it into your job template. Each job temp
...
+```
+
+SSL Issues
+----------
+
+If you don't have the CA that secures our Jenkins instance in your system CA pack JJB will complain
+about SSL issues. To fix this add the CA to your system CA pack. Alternatively you can modify the
+JJB check to allow insecure usage. Warning: allow ***insecure at your own risk***.
+
+To workaround SSL errors patch line 431 of `/usr/lib/python2.7/site-packages/jenkins/__init__.py`
```
Ichimonji10: I know there's an "at your own risk" disclaimer right above this line, but... should we really be advising that people patch JJB's system-wide source code to allow insecure connections? There's got to be a better way.

ci/jobs/README.md
Ichimonji10: OK. Given that this URL is valid, is the password on the line above valid?

bmbouter: No the password is fake.

ci/jobs/README.md
bmbouter: This hack is so dirty I almost feel bad publishing it. I also want a better way. I wish JJB had an `--insecure` option. What I think would be *really* sweet is if we got our Jenkins public and used `jenkins.pulpproject.org` and used a CA from a typically trusted CA so we wouldn't even need this workaround. For now I want to share the notes I already have with others.

PR #372, opened 2017-05-30
Pulp smash new settings

elyezer:
Added two commits to:

1. Update Pulp Smash ansible role to use the new Pulp Smash settings file format. Had to rename variables to be more meaningful.
2. Update all the Pulp automation jobs to properly trigger the updated pulp-smash-runner job which is now accepting the parameters to deal with the new Pulp Smash settings file format.


elyezer:
If I don't get any review by tomorrow morning I am going to merge this so I can finally work on the clustered Pulp jobs.

PR #374, opened 2017-06-03
Change BZ status to POST if the redmine bug is in CLOSED - COMPLETE

goosemania:


goosemania:
Example of the issue which does not cause BZ to change its status: https://pulp.plan.io/issues/2638

PR #380, opened 2017-06-08
Update satellite6-upgrade-pulp jobs

elyezer:
Do the following:

* Upgrade Pulp to 2.13 on downstream Satellite
* Drop upstream Satellite job since it is already packet with 2.13
* Use openstack client since nova client has deprecated some of its
subcommands in favor of the openstack client.
* Install yum-utils before running yum-config-manager


elyezer:
The current changes are already in place on downstream-satellite6-upgrade-pulp job and it got a green build. :tada:

PR #383, opened 2017-07-03
Stop setting DJANGO_SETTINGS_MODULE in the doc builder

werwty:
closes #2819
https://pulp.plan.io/issues/2819

PR #389, opened 2017-07-06
Update doc builder jobs to not set DJANGO_SETTINGS_MODULE

werwty:
closes #2819
https://pulp.plan.io/issues/2819


werwty:
I accidentally updated my branch before merging https://github.com/pulp/pulp_packaging/pull/383 so the change never made it in. But since the same change was already approved, I'm merging this without additional approval.

PR #393, opened 2017-07-13
fix pulp install on machines with apache 2.2 and iptables

kdelee:
I have tested running the pulp_server.yaml playbook to install pulp on a clean RHEL6.9 server and RHEL7.3 server (my F25 laptop as control node)

I did move around some existing tasks into blocks since more tasks that all were conditional on `(ansible_distribution == "RedHat" and ansible_distribution_major_version|int == 6)` were added.

@elyezer what more should I do to get this ready to merge?

Thanks @Ichimonji10 for giving me a head start on this fix

### DIFF
--- a/ci/ansible/roles/pulp-crane/files/pulp_crane_el6.conf
+++ b/ci/ansible/roles/pulp-crane/files/pulp_crane_el6.conf
@@ -1,0 +1,15 @@
+# source:
+# https://raw.githubusercontent.com/pulp/crane/3c657fe53020b804ce60d907a82b4fb19aebe532/deployment/apache22.conf
+# Use this config with Apache 2.2 or earlier
+Listen 5000
+<VirtualHost *:5000>
+    WSGIScriptAlias / /usr/share/crane/crane_el6.wsgi
+    <Location /crane>
+        Order deny,allow
+        Allow from all
+    </Location>
+    <Directory /usr/share/crane/>
+        Order allow,deny
+        Allow from all
+    </Directory>
+</VirtualHost>
--- a/ci/ansible/roles/pulp-crane/tasks/main.yml
+++ b/ci/ansible/roles/pulp-crane/tasks/main.yml
@@ -30,9 +30,25 @@
     group: root
     mode: 0644
 
-- name: Open firewall port for Crane
+- name: Open firewall port for Crane (iptables)
+  iptables:
+    action: insert
+    chain: INPUT
+    ctstate: NEW
+    protocol: tcp
+    destination_port: 5000
+    jump: ACCEPT
+  when:
+    - ansible_distribution == "RedHat"
+    - ansible_distribution_major_version|int == 6
+  notify:
+    - Save IPv4 iptables configuration
+
+- name: Open firewall port for Crane (firewalld)
   firewalld:
     port: "5000/tcp"
     state: enabled
     permanent: true
     immediate: true
+  when:
+    - not (ansible_distribution == "RedHat" and ansible_distribution_major_version|int == 6)
--- a/ci/ansible/roles/pulp-crane/tasks/main.yml
+++ b/ci/ansible/roles/pulp-crane/tasks/main.yml
@@ -13,6 +13,19 @@
     group: root
     mode: 0644
   notify: restart apache
+  when:
+    - not (ansible_distribution == "RedHat" and ansible_distribution_major_version|int == 6)
+
+- name: Install Apache configuration file for Apache version 2.2 or less
+  copy:
+    src: pulp_crane_el6.conf
+    dest: /etc/httpd/conf.d/pulp_crane_el6.conf
+    owner: root
+    group: root
+    mode: 0644
+  notify: restart apache
+  when:
+    - (ansible_distribution == "RedHat" and ansible_distribution_major_version|int == 6)
 
 # NOTE: This task exists as a work-around for https://pulp.plan.io/issues/2719
 # When that issue is fixed, this task should be removed.
@@ -30,20 +43,6 @@
     group: root
     mode: 0644
 
-- name: Open firewall port for Crane (iptables)
-  iptables:
-    action: insert
-    chain: INPUT
-    ctstate: NEW
-    protocol: tcp
-    destination_port: 5000
-    jump: ACCEPT
-  when:
-    - ansible_distribution == "RedHat"
-    - ansible_distribution_major_version|int == 6
-  notify:
-    - Save IPv4 iptables configuration
-
 - name: Open firewall port for Crane (firewalld)
   firewalld:
     port: "5000/tcp"
@@ -51,4 +50,26 @@
     permanent: true
     immediate: true
   when:
-    - not (ansible_distribution == "RedHat" and ansible_distribution_major_version|int == 6)
+  - not (ansible_distribution == "RedHat" and ansible_distribution_major_version|int == 6)
+
+- block:
+  - name: Install Crane WSGI config file (RHEL 6)
+    copy:
+      src: crane_el6.wsgi
+      dest: /usr/share/crane/crane_el6.wsgi
+      owner: root
+      group: root
+      mode: 0644
+
+  - name: Open firewall port for Crane (iptables)
+    iptables:
+      action: insert
+      chain: INPUT
+      ctstate: NEW
+      protocol: tcp
+      destination_port: 5000
+      jump: ACCEPT
+    notify:
+      - Save IPv4 iptables configuration
+  when:
+    - (ansible_distribution == "RedHat" and ansible_distribution_major_version|int == 6)
--- a/ci/ansible/roles/pulp/tasks/main.yaml
+++ b/ci/ansible/roles/pulp/tasks/main.yaml
@@ -1,23 +1,38 @@
 ---
+- block:
+  - name: Open firewall ports (iptables)
+    iptables:
+      action: insert
+      chain: INPUT
+      ctstate: NEW
+      protocol: "{{ item.protocol }}"
+      destination_port: "{{ item.port }}"
+      jump: ACCEPT
+    with_items:
+      - { port: 'http', protocol: 'tcp' }
+      - { port: 'https', protocol: 'tcp' }
+      - { port: 'amqps', protocol: 'tcp' }
+      - { port: 'amqp', protocol: 'tcp' }
+    notify:
+      - Save IPv4 iptables configuration
 
-- name: Open firewall ports (iptables)
-  iptables:
-    action: insert
-    chain: INPUT
-    ctstate: NEW
-    protocol: "{{ item.protocol }}"
-    destination_port: "{{ item.port }}"
-    jump: ACCEPT
-  with_items:
-    - { port: 'http', protocol: 'tcp' }
-    - { port: 'https', protocol: 'tcp' }
-    - { port: 'amqps', protocol: 'tcp' }
-    - { port: 'amqp', protocol: 'tcp' }
+  - name: Install required packages for RHEL 6
+    action: "{{ ansible_pkg_mgr }} name={{ item }} state=latest"
+    with_items:
+        - libselinux-python
+        - policycoreutils-python
+
+  - name: Setup qpid custom repo
+    get_url: url=https://copr.fedorainfracloud.org/coprs/g/qpid/qpid/repo/epel-6/irina-qpid-epel-6.repo dest=/etc/yum.repos.d/copr-qpid.repo
+
+  - name:  Allow Apache to listen on tcp port 5000
+    seport:
+      ports: 5000
+      proto: tcp
+      setype: http_port_t
+      state: present
   when:
-    - ansible_distribution == "RedHat"
-    - ansible_distribution_major_version|int == 6
-  notify:
-    - Save IPv4 iptables configuration
+      - (ansible_distribution == "RedHat" and ansible_distribution_major_version|int == 6)
 
 - block:
   - name: Install firewalld and python bindings
@@ -56,12 +71,6 @@
 
 - name: Start and enable MongoDB server service
   service: name=mongod state=started enabled=yes
-
-- name: Setup qpid custom repo
-  get_url: url=https://copr.fedorainfracloud.org/coprs/g/qpid/qpid/repo/epel-6/irina-qpid-epel-6.repo dest=/etc/yum.repos.d/copr-qpid.repo
-  when:
-    - ansible_os_family == "RedHat"
-    - ansible_distribution_major_version|int == 6
 
 - name: Install qpid-cpp server
   action: "{{ ansible_pkg_mgr }} name={{ item }} state=latest"
--- a/ci/ansible/roles/pulp-crane/files/crane_el6.wsgi
+++ b/ci/ansible/roles/pulp-crane/files/crane_el6.wsgi
@@ -1,0 +1,7 @@
+# see:
+# https://raw.githubusercontent.com/pulp/crane/3c657fe53020b804ce60d907a82b4fb19aebe532/deployment/crane_el6.wsgi
+# for this unpleasantry is necessary.
+import sys
+sys.path.insert(0, '/usr/lib/python2.6/site-packages/Jinja2-2.6-py2.6.egg')
+
+from crane.wsgi import application
--- a/ci/ansible/roles/pulp-crane/handlers/main.yml
+++ b/ci/ansible/roles/pulp-crane/handlers/main.yml
@@ -2,3 +2,6 @@
   service:
     name: httpd
     state: restarted
+
+- name: Save IPv4 iptables configuration
+  shell: 'iptables-save > /etc/sysconfig/iptables'


ci/ansible/roles/pulp-crane/tasks/main.yml
```diff
@@ -22,17 +35,42 @@
...
+      mode: 0644
+
+  - name: Open firewall port for Crane (firewalld)
+    firewalld:
+      port: "5000/tcp"
+      state: enabled
+      permanent: true
+      immediate: true
+  when:
+    - not (ansible_distribution == "RedHat" and ansible_distribution_major_version|int == 6)
```
elyezer: I would normalize conditional blocks to use the same logic. For example on line 17 it is checking if the major_version is greater than here it is not equal. That said I would normalize and use greater than approach or not equal on both places.

Also, if choosing the not equal approach maybe `ansible_distribution == "RedHat" and ansible_distribution_major_version|int != 6` is more readable.

kdelee: Hmm, you bringing this up makes me realize that perhaps the other one should be the same as this, because 
"not (this AND that)" == "(not this OR not that)" and I'm not sure the other line actually will do the right thing on Fedora. I'll make it consistent and it and test it out on Fedora 25.

ci/ansible/roles/pulp-crane/files/pulp_crane_el6.conf
```diff
@@ -0,0 +1,32 @@
...
+        Deny from all
+        Allow from localhost
+    </Location>
+    <Directory /usr/share/crane/>
+        Order allow,deny
+        Allow from all
+    </Directory>
+</VirtualHost>
+
+# <VirtualHost *:443>
```
Ichimonji10: Can the commented block be dropped?

ci/ansible/roles/pulp-crane/tasks/main.yml
Ichimonji10: > Also, if choosing the not equal approach maybe ansible_distribution == "RedHat" and ansible_distribution_major_version|int != 6 is more readable.

That would break this playbook on other distributions like CentOS 7. I think the best approach is to assume certain things by default, like that systemd and firewalld are in use, and then make exceptions for distros like RHEL 6 as necessary.

ci/ansible/roles/pulp-crane/files/pulp_crane_el6.conf
kdelee: I'm fine with that. i just grabbed this file from pulp crane:
https://github.com/pulp/crane/blob/master/deployment/apache22.conf

Ichimonji10: :+1:

ci/ansible/roles/pulp-crane/tasks/main.yml
kdelee: @Ichimonji10 :  I'm not clear on what should be done based on this exchange.

kdelee:
@Ichimonji10 @elyezer 

I made some changes and tested on Fedora25 Server, RHEL 6.9 server, and RHEL 7.3 server.

Then ran pulp_smash remotely on each machine (just pulp_smash.tests.platform.api_v2)
All was well.

Let me know what y'all want to do about guarding the firewalld and such tasks

ci/ansible/roles/pulp-crane/tasks/main.yml
Ichimonji10: I'm advocating for conditions like:

```yaml
# RHEL 6
when:
  - ansible_distribution == "RedHat" and ansible_distribution_major_version == 6

# non-RHEL 6
when:
  - not (ansible_distribution == "RedHat" and ansible_distribution_major_version == 6)
```

One nice property of using this pair of conditions it covers every system we might target. For example, what if we target a CentOS 7 system? Or an OpenSUSE system? They'll both count as "non-RHEL 6", and the latter block of code will run.

On the other hand, this pair of conditions doesn't cover every system we might target:

```yaml
# RHEL 6
when:
  - ansible_distribution == "RedHat" and ansible_distribution_major_version == 6

# All of RHEL/Fedora/CentOS except RHEL 6, and not OpenSuse etc
when:
  - ansible_distribution == "RedHat" and ansible_distribution_major_version != 6
```

kdelee: Ok, well right now I've gone with the first option, which you seem to prefer. 😃

ci/ansible/roles/pulp/tasks/main.yaml
```diff
@@ -1,23 +1,40 @@
...
+    get_url: url=https://copr.fedorainfracloud.org/coprs/g/qpid/qpid/repo/epel-6/irina-qpid-epel-6.repo dest=/etc/yum.repos.d/copr-qpid.repo
+
+  - name:  Allow Apache to listen on tcp port 5000
+    seport:
+      ports: 5000
+      proto: tcp
+      setype: http_port_t
+      state: present
+    notify:
+      - Save IPv4 iptables configuration
```
Ichimonji10: Does the `seport` module fiddle with iptables rules?

What makes this module specific to RHEL 6?

ci/ansible/roles/pulp/tasks/main.yaml
```diff
@@ -1,23 +1,40 @@
...
-    - { port: 'https', protocol: 'tcp' }
-    - { port: 'amqps', protocol: 'tcp' }
-    - { port: 'amqp', protocol: 'tcp' }
+  - name: Install required packages for RHEL 6
+    action: "{{ ansible_pkg_mgr }} name={{ item }} state=latest"
+    with_items:
+        - libselinux-python
+        - policycoreutils-python
+    notify:
+      - Save IPv4 iptables configuration
```
Ichimonji10: Why is this handler being notified?

Ichimonji10:
I'm still reviewing this PR in full. I've just commented on the things I've noticed right off the bat.

ci/ansible/roles/pulp-crane/tasks/main.yml
```diff
@@ -13,6 +13,19 @@
     group: root
     mode: 0644
   notify: restart apache
+  when:
+    - not (ansible_distribution == "RedHat" and ansible_distribution_major_version|int == 6)
+
+- name: Install Apache configuration file for Apache version 2.2 or less
+  copy:
+    src: pulp_crane.conf
```
Ichimonji10: Shouldn't `src` b e `pulp_crane_el6.conf` ?

ci/ansible/roles/pulp/tasks/main.yaml
kdelee: good call

ci/ansible/roles/pulp-crane/files/pulp_crane_el6.conf
```diff
@@ -0,0 +1,15 @@
+# source:
+# https://raw.githubusercontent.com/pulp/crane/3c657fe53020b804ce60d907a82b4fb19aebe532/deployment/apache22.conf
+# Use this config with Apache 2.2 or earlier
+<VirtualHost *>
```
Ichimonji10: Did you test whether Pulp+Crane works when accessed from a Docker client? I strongly suspect that this config file is broken, for a couple reasons:

* This `VirtualHost` directive listens on port 80, not port 5000.
* The `Location` directive listens only for Docker requests that originate from localhost.

The sample Apache 2.4 configuration file provided by the Crane repository is basically broken for production purposes by default, and we don't use it here in pulp_packaging. The same may be true for the sample Apache 2.2 configuration file.

ci/ansible/roles/pulp/tasks/main.yaml
kdelee: The way the SELinux settings were set on the RHEL6 machines I was getting from beaker, Apache was not being allowed to bind to the port. This was the solution that I found. 

For reference, this seems to be a issue other RHEL6 users find:
[rhel6 allow httpd to connect to specific port](https://serverfault.com/questions/563872/selinux-allow-httpd-to-connect-to-a-specific-port)

I followed the example on the seport module and it worked, so I was satisfied.
[examples](http://docs.ansible.com/ansible/seport_module.html#examples)

I needed it for RHEL 6, but it was previously unnecessary, so that's why I guarded it in this block.
Honestly, I think the notify statement being there is a mistake. It should be up under the iptables task and it got pushed down there when I was re-organizing the tasks into blocks.

ci/ansible/roles/pulp-crane/files/pulp_crane_el6.conf
kdelee: Can you direct me to something (docs, pastebin of some code/psuedo code, something else) that I could do to test it?

ci/ansible/roles/pulp/tasks/main.yaml
Ichimonji10: > The way the SELinux settings were set on the RHEL6 machines I was getting from beaker, Apache was not being allowed to bind to the port.

Word. Makes sense.

kdelee:
@Ichimonji10 

Thanks for looking at this. I can test it this afternoon again. I will look into how to confirm that pulp-crane is working. If there are some pulp-smash tests you can suggest, that would be great.

Ichimonji10:
Smash tests are a good place to start:

    python -m unittest discover pulp_smash.tests.docker

The tests kinda suck, though, because they don't use the real Docker client. I'll post an example of how to run a manual test with the real Docker client.

Ichimonji10:
A basic manual test is as follows:

1. Create a Docker repository with a (v2) feed of https://registry-1.docker.io.
2. Sync the library/busybox image.
3. Publish the repository.
4. Run `docker pull localhost:5000/repo-name`.
5. Execute `docker run localhost:5000/repo-name echo foo` and `docker run image-checksum echo foo`.

A Docker repository can be created with a command like:

```bash
pulp-admin docker repo create \
  --feed https://registry-1.docker.io \
  --upstream-name busybox \
  --repo-id foo \
  --auto-publish false
```

kdelee:
@Ichimonji10 
Great, thanks I appreciate your help with this. I also talked with @dkliban some about crane and he said he could help me if I continue to run into trouble.

I'll let you know how it goes.

kdelee:
@Ichimonji10 
Ok, well, I'm passing the `pulp_smash.tests.docker` tests but for the life of me I don't know how to form the URL to tell my local docker client to pull it down from the remote host (docker packages are not available for that version of RHEL)

Ichimonji10:
> I get an error saying `server gave HTTP response to HTTPS client`

I think that's because our Ansible playbook for Crane makes crane use HTTP, not HTTPS.  I need to look at the Crane documentation to see exactly which bit of configuration that is.

ci/ansible/roles/pulp-crane/files/pulp_crane_el6.conf
```diff
@@ -0,0 +1,16 @@
+# source:
+# https://raw.githubusercontent.com/pulp/crane/3c657fe53020b804ce60d907a82b4fb19aebe532/deployment/apache22.conf
+# Use this config with Apache 2.2 or earlier
+Listen 5000
+<VirtualHost *:5000>
+    WSGIScriptAlias / /usr/share/crane/crane_el6.wsgi
+    <Location /crane>
+        Order Deny,Allow
+        Deny from all
+        Allow from localhost
```
Ichimonji10: For testing purposes, I believe connections to Crane should be allowed from clients other than localhost.

kdelee:
Ok, so, if I allow docker to make HTTP connections on my machine:
```
 $  cat /etc/docker/daemon.json 
{
  "insecure-registries" : ["${PULPHOST}:5000"]
}
```

AND I make sure that the wsgi file is in `/usr/share/crane_el6.wsgi` as is pointed to in `/etc/httpd/conf.d/pulp_crane_el6.conf` NOW I seem to be able to make the connection with docker.

I am getting a failure:
```
docker pull ${PULPHOST}:5000/busybox:latest
Trying to pull repository ibm-x3550m3-12.lab.eng.brq.redhat.com:5000/busybox ... 
sha256:74155e76e3396bc23d5b8cedf606bc04f7f79660ee5bce14014b84651fc4f672: Pulling from ibm-x3550m3-12.lab.eng.brq.redhat.com:5000/busybox

27144aa8f1b9: Verifying Checksum 
filesystem layer verification failed for digest sha256:27144aa8f1b9e066514d7f765909367584e552915d0d4bc2f5b7438ba7d1033a
```

But given how much I've been fiddling around I'm going to try from a fresh start and make sure ansible is putting the files in the right place now that I've 'corrected' my previous problems.

kdelee:
@Ichimonji10 
OK, so I just ran the pulp_server playbook on a RHEL6.9 machine and once I had created/synced/published a docker repo on it and added the host to my `/etc/docker/daemon.json ` on my machine, I again get this checksum verification error. @dkliban has come to the conclusion that I am setting it up correctly, and that this feature is just broken on RHEL6 because of https://pulp.plan.io/issues/2618.

I made a story on redmine to track my work on this project, as it spans beyond this PR:
[As the admin user for a Pulp 2.11.z server running on RHEL 6.9, I am able to backup my database and import into a new Pulp 2.13.z server running on RHEL 7.3+](https://pulp.plan.io/issues/2901)

ci/ansible/roles/pulp-crane/tasks/main.yml
```diff
@@ -30,9 +43,33 @@
     group: root
     mode: 0644
```
Ichimonji10: This "Install Crane configuration file" task should now have a `when:` clause. Without such a clause, this task will always execute. On RHEL 6, the "Install Crane configuration file (RHEL 6)" task will also execute, causing two unnecessary modifications on every run.

kdelee: I think that this may be a case of bad task names. Both systems get the SAME file (crane.conf)  in /etc/crane.conf
```
[general]
data_dir: /var/lib/pulp/published/docker
debug: false
```
Which is put in place by the `Install Crane configuration file` task

Only el6 gets `crane_el6.wsgi` in `/usr/share/crane/crane_el6.wsgi` which has the task name `Install Crane configuration file (RHEL 6)`, but it is totally different/has different purpose.  I'll change it to `Install Crane WSGI config file (RHEL6)`.
```
sys.path.insert(0, '/usr/lib/python2.6/site-packages/Jinja2-2.6-py2.6.egg')

from crane.wsgi import application
```

kdelee: @Ichimonji10 
Ok I changed the task name to disambiguate. Does that make more sense now? EL6 needs an additional config file.

Ichimonji10: Sorry about the mix-up. I mis-read the code.

ci/ansible/roles/pulp-crane/files/pulp_crane_el6.conf
```diff
@@ -0,0 +1,16 @@
+# source:
+# https://raw.githubusercontent.com/pulp/crane/3c657fe53020b804ce60d907a82b4fb19aebe532/deployment/apache22.conf
+# Use this config with Apache 2.2 or earlier
+Listen 5000
+<VirtualHost *:5000>
+    WSGIScriptAlias / /usr/share/crane/crane_el6.wsgi
+    <Location /crane>
+        Order deny,allow
+        Deny from all
+        Allow from localhost
```
Ichimonji10: This `<Location>` blocks network access to all `/crane/*` URLs by hosts other than localhost. The following should fix the issue:

```apache
<Location /crane>
Order deny,allow
Allow from all
</Location>
```

There's probably some more elegant fix. It might be possible to drop the `<Location>` block altogether.

Ichimonji10: The change I suggested above works. You can test for yourself by executing this on the Pulp+Crane system:

```sh
curl -H accept:application/json http://localhost:5000/crane/repositories
```

...and this from some other host:

```sh
curl -H accept:application/json http://${hostname}:5000/crane/repositories
```

Before the change, the latter request results in an HTTP 403 "forbidden" response. After the change, the request is answered.

kdelee: That's interesting, I make contact with a remote rhel6 machine configured as-is with a docker client on my machine.

Maybe the docker client itself only ends up trying to get data in `/usr/share/crane` ?

If this change is important for testing, please go ahead and push it and I can spark the jenkins jobs testing this branch (they build with it and then run pulp smash)

kdelee: @Ichimonji10 
Ok, tried it out, it works for me too!

Ichimonji10: > Maybe the docker client itself only ends up trying to get data in `/usr/share/crane` ?

Aye. It's probably the case that Docker doesn't need to access the `/crane/repositories` path, so I over-stated the importance of this change in that regard.

The path is super-helpful in tests. The tests in `pulp_smash.tests.docker.cli.test_sync_publish` do use paths like `/crane/repositories`.

PR #395, opened 2017-07-15
Add supported version warning to docs

JayWelborn:
Added warning message to docs as per https://pulp.plan.io/issues/2673

This will be accompanied by a PR on pulp-2.13-dev to update build docs so that changing the 'supported-releases.json' file is included each time a new version is released.


ci/config/releases/supported-releases.json
```diff
@@ -0,0 +1,3 @@
+[ 
+    "2.13",
```
werwty: comma needs to be removed for this to be valid json

ci/docs/layout.html
```diff
@@ -12,4 +12,36 @@
...
+{# check current version against supported-releases.json #}
+<script>
+$( document ).ready( function() {
+
+  // gets version number minus whitespace and bugfix version
+  var $version_number = $('.version').text().trim().split('.', 2).join('.')
+
+  var supported_versions = []
+  // get JSON from server and load supported_versions
+  $.getJSON("https://raw.githubusercontent.com/pulp/pulp_packaging/master/ci/config/releases/supported-releases.json" , function(data) {
```
werwty: Since this is an asynchronous request supported_versions is not guaranteed to be populated when ``if ($.inArray($version_number, supported_versions) == -1)`` is being checked. 
Consider adding the check in a .done() callback function like this: http://api.jquery.com/jquery.getjson/#jqxhr-object

ci/docs/layout.html
```diff
@@ -12,4 +12,36 @@
...
+    $.each(data, function(value) {
+      supported_versions.push(value)
+    })
+  })
+
+  console.log(supported_versions)
+
+  // render warning if current version not in supported_versions
+  if ($.inArray($version_number, supported_versions) == -1) {
+    var warning = $('\
```
werwty: warning can be just a string and not a jquery object.

ci/docs/layout.html
```diff
@@ -12,4 +12,36 @@
...
+        <div class="admonition warning"> \
+            <p class="first admonition-title">Warning</p> \
+            <p class="last">You are looking at the docs for an unsupported version of Pulp. \
+             Consider upgrading to a  \
```
werwty: s/upgrading/switching/

Since the latest nightly doc builds are not "supported", the verb "upgrading" is not the best choice.

ci/docs/layout.html
JayWelborn: This is actually my first time using ajax. Would the following be valid?


`var ajax = $.getJSON( //stuff)`
`ajax.done(function() { //other stuff }`

werwty: I would rewrite this to be: 

```
$( document ).ready( function() {
	var $version_number = $('.version').text().trim().split('.', 2).join('.')
	
	// Declared outside of the ajax promise since it does not change
	var warning = '\
        <div class="admonition warning"> \
            <p class="first admonition-title">Warning</p> \
            <p class="last">You are looking at the docs for an unsupported version of Pulp. \
             Consider upgrading to a  \
             <a href="http://pulpproject.org/docs/">supported version</a></p> \
        </div>'

	// this is not necessary since the returned value is already an array
	// var supported_versions = []


	$.getJSON("https://raw.githubusercontent.com/pulp/pulp_packaging/master/ci/config/releases/supported-releases.json" , function(data) {
	// this is technically the success callback so no need for $.done() (I need to read docs more carefully)
		if ($.inArray($version_number, data) == -1) {
			$('.rst-content').prepend(warning)
		}
	})
	
})
```

ci/docs/layout.html
```diff
@@ -12,4 +12,36 @@
...
+  $.getJSON("https://raw.githubusercontent.com/pulp/pulp_packaging/master/ci/config/releases/supported-releases.json" , function(data) {
+    $.each(data, function(value) {
+      supported_versions.push(value)
+    })
+  })
+
+  console.log(supported_versions)
+
+  // render warning if current version not in supported_versions
+  if ($.inArray($version_number, supported_versions) == -1) {
```
werwty: this should be === because javascript. (it probably doesn't matter here, but it might because js sometimes does some questionable things)

ci/docs/layout.html
JayWelborn: The warning declaration outside the getJSON call and the append within the success callback makes a lot of sense. New PR incoming today with these changes.

Thanks for the help!

JayWelborn:
Last commits incorporates requested changes. Should I squash these last three into one?

Ichimonji10:
> Should I squash these three into one?

Yes.

ci/config/releases/supported-releases.json
```diff
@@ -0,0 +1,3 @@
+[ 
+    "2.12"
```
werwty: Could you change this back to 2.13?

JayWelborn:
fixed and squashed!

PR #396, opened 2017-07-16
Add Plugins section, manage plugins/index.rst differently for Pulp 3

goosemania:
re #2852
https://pulp.plan.io/issues/2852

### DIFF
--- a/ci/docs-builder.py
+++ b/ci/docs-builder.py
@@ -77,24 +77,26 @@ def main():
         dst = os.sep.join([plugins_dir, component['name']])
         os.symlink(src, dst)
 
-    # copy in the pulp_index.rst file
     if is_pulp3:
-        src_path = 'docs/pulp_index_pulp3.rst'
+        src_index_path = 'docs/pulp_index_pulp3.rst'
+        src_all_content_path = 'docs/all_content_index_pulp3.rst'
     else:
-        src_path = 'docs/pulp_index.rst'
-    pulp_index_rst = os.sep.join([WORKING_DIR, 'pulp', 'docs', 'index.rst'])
-    copyfile(src_path, pulp_index_rst)
+        src_index_path = 'docs/pulp_index.rst'
+        src_all_content_path = 'docs/all_content_index.rst'
 
-    # copy in the plugin_index.rst file
-    plugin_index_rst = os.sep.join([plugins_dir, 'index.rst'])
-    copyfile('docs/plugin_index.rst', plugin_index_rst)
+        # copy in the plugin_index.rst file for Pulp 2 only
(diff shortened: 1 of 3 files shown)


ci/docs-builder.py
```diff
@@ -77,24 +77,26 @@ def main():
...
-    copyfile(src_path, pulp_index_rst)
+        src_index_path = 'docs/pulp_index.rst'
+        src_all_content_path = 'docs/all_content_index.rst'
 
-    # copy in the plugin_index.rst file
-    plugin_index_rst = os.sep.join([plugins_dir, 'index.rst'])
-    copyfile('docs/plugin_index.rst', plugin_index_rst)
+        # copy in the plugin_index.rst file for Pulp 2 only
+        # (currently Pulp 3 has its own plugins/index.rst without a need of managing it here,
+        # outside of platform code)
```
goosemania: I am not sure if it's a right solution. Maybe it's better to create a separate `plugin_index_pulp3.rst` file here and manage it similar to Pulp 2 docs.

werwty: I think we'll need a plugin_index_pulp3.rst later when we start building plugin docs. For not it's not necessary

ci/docs/pulp_index_pulp3.rst
```diff
@@ -42,4 +41,4 @@ Index and Search
 .. toctree::
    :maxdepth: 1
 
-   all_content_index_pulp3
+   all_content_index
```
werwty: why not all_content_index_pulp3 here?

goosemania: @werwty , yeah, that's a confusing part which I got wrong last time...
No matter what the filename in this for the all_content_index is, [it will be copied and become an `all_content_index.rst`](https://github.com/goosemania/pulp_packaging/blob/2aa8032bb18759d09a9a1e2b9e8aa82d3bbef3c6/ci/docs-builder.py#L97-L99). So `all_content_index` in a toctree is correct for any Pulp version.

goosemania: Does it make sense or just more confusing? :)

I also suggest to postpone review of this PR just  a bit, while I am trying to make a proposal for building docs. And this particular piece of code may change.

werwty: @goosemania oh I gotcha. 
I reviewed this because parts of it should be merged with you pulp plugin docs pr right?

I think this is good to merge as is if it blocks getting that merged. We can always change this code later

goosemania: @werwty , yeah, sure we can merge that and refactor later

PR #398, opened 2017-07-18
Replace deprecated job property: logrotate

Ichimonji10:
The `logrotate` job property is deprecated as of Jenkins 1.637. Use the
`build-discarder` job property instead. See:

* https://docs.openstack.org/infra/jenkins-job-builder/definition.html#job
* https://docs.openstack.org/infra/jenkins-job-builder/properties.html#properties.build-discarder


Ichimonji10:
Our Jenkins instance is running Jenkins 1.651.3.

Tested by running JJB against a personal instance of Jenkins 2.70.

PR #400, opened 2017-07-18
Update jenkins jobs for 2.14

pcreech:
Update jenkins jobs for 2.14


werwty:
I think you need to add 2.14 in a couple more places:
https://github.com/pulp/pulp_packaging/blob/master/ci/docs-builder.py#L12
https://github.com/pulp/pulp_packaging/blob/master/ci/jobs/pr-docs.yaml#L46
https://github.com/pulp/pulp_packaging/blob/master/ci/jobs/pulp-installer.yaml#L24

PR #402, opened 2017-07-19
Add plugins to the Pulp docs according to the release config

goosemania:
For any component 'build_docs' option can be specified
in the release config. If disabled, it allows not to build docs
for plugins if not needed (e.g. community plugin hosts their
docs somewhere else).

Index files for platform docs are moving to the platform repository.
Plugins' index file will be copied in as before but unnecessary
plugins will be filtered out from it.


goosemania:
That's a proposal. Additional PRs needed to move some index files to the platform repo.

bmbouter:
Here is my super stupid fix of the immediate problem.  https://github.com/pulp/pulp_packaging/pull/407

bmbouter:
The other PR is merged and the the `2.14-dev` and `master` docs passed and published. The 2.14-build docs failed though due to pulp_docker so we need to fix those somehow.

bmbouter:
pulp_docker got fixed this morning, so the docs are all building successfully now. I think this can be closed.

PR #405, opened 2017-07-21
Reorganize JJB-related data

Ichimonji10:
Move JJB-related data from `ci/jobs` to `ci/jjb`. This change more
accurately reflects what's actually in this directory: not just jobs,
but also scripts, configuration files, a readme and a job template.

Move JJB job definitions from `ci/jjb` to `ci/jjb/jobs`. This change
allows for better organization. The readme, sample configuration file
and job template reside in `ci/jjb`, job definitions reside in
`ci/jjb/jobs`, and scripts reside in `ci/jjb/scripts`.

Rename the sample configuration file from `jenkins_jobs.ini.example` to
`jenkins_jobs.ini`. Letting the file's suffix be `.ini` lets text
editors and other tools easily recognize its format. Rename the template
from `template.yaml.sample` to `template.yaml`, both for the just-stated
reason, and also because the template is no longer in the same directory
as all of the real job definitions. (Hooray for organization.)

Encourage the user to do all their work in the `ci/jjb` directory. A
typical workflow might be:

```sh
cd ci/jjb
less README.md  # learn following steps
install -Dm600 jenkins_jobs.ini ~/.config/jenkins_jobs/jenkins_jobs.ini
vim ~/.config/jenkins_jobs/jenkins_jobs.ini
jenkins-jobs test jobs
```

There are fewer steps than in the previous suggested workflow:

```sh
cd ci/jobs
less README.md  # learn following steps
mkdir ~/.config/jenkins_jobs/
cp jenkins_jobs.ini.sample ~/.config/jenkins_jobs/jenkins_jobs.ini
vim ~/.config/jenkins_jobs/jenkins_jobs.ini
cd ../../
jenkins-jobs --include-path ci/jobs/scripts test ci/jobs
```

Update the `jenkins_jobs.ini` configuration file:

* Drop the `allow_duplicates=False` directive. This option already
  defaults to false.
* Add an `ignore_cache=True` directive. The readme formerly suggested
  that this option be passed as a CLI option.
* Update the `include_path` directive to account for path changes.

Reformat the JJB readme. Hard-wrap all text at 80 columns. (It formerly
had a mixture of widths.) Fix some obvious semantic errors, especially
with mis-usages of inline code and block code markers. Fix typos. Change
some awkward language. Account for all the other changes described above.

Update the `.travis.yml` file to account for the changes described
above.


Ichimonji10:
The `ci/jjb/jobs/ci-update-jobs.yaml` file has not been updated. That job was already broken, and it's now even more broken.

elyezer:
> The ci/jjb/jobs/ci-update-jobs.yaml file has not been updated. That job was already broken, and it's now even more broken.

We can remove that job

Ichimonji10:
@dkliban @bmbouter @pcreech Thoughts?

dkliban:
This seems like a reasonable change to me. I just checked and I didn't see any job definitions that rely on this directory structure. I'd also like to hear from @bmbouter

ci/jjb/README.md
```diff
@@ -0,0 +1,142 @@
...
+documentation.
+
+JJB Usage
+---------
+
+Two of the most important things one can do with JJB are to generate jobs and
+print them to stdout, and to generate jobs and send them to a Jenkins instance:
+
+```sh
+jenkins-jobs test jobs'
```
bmbouter: Is the apostrophe at the end of this and the next line intended?

elyezer: good catch, I think they are not needed.

Ichimonji10: Whoops!

Ichimonji10:
> Something to think about, we could move the jjb from within ci dir to the repository root dir. I mean we can do that for everything inside ci dir since that level of organization is just an additional dir level without a really good benefit.

> We can remove [the `ci-update-jobs.yaml`] job

I agree. I'm going to leave this PR as-is, so as to avoid changing things that people have already reviewed. But these changes would be nice.

elyezer:
Sounds good, I think we can merge this. Thank you :ship: it

PR #412, opened 2017-08-01
Moves around the owner of two jobs

bmbouter:
* Rename the bz-automation-ownership group to be the
  satellite-ownership group. Also update existing jobs to use the new
  name.
* Move the pr-docs builder to be owned by the docs-ownership group
* Move the rebase_analysis job to be owned by the satellite-ownership
  group.


kdelee:
LGTM 👍

PR #414, opened 2017-08-07
Work around inability to install groups on F26

Ichimonji10:
Ansible is currently unable to install package groups on Fedora 26. Work
around this limitation. See:
https://github.com/ansible/ansible/issues/26868


Ichimonji10:
It's not yet possible to install Pulp 2.14 on Fedora 26 with our Ansible playbooks. Fixing this issue reveals an issue with the SSL configuration procedure. But this gets us closer.

PR #415, opened 2017-08-07
Ensure /etc/pki/CA is present on Fedora 26

Ichimonji10:
The pulp-certs Ansible role requires that `/etc/pki/CA` directory be
present.


Ichimonji10:
#414 and #415 must both be merged for Pulp to successfully be installed. Unfortunately, even with both of these PRs merged, pulp_packaging doesn't produce a working Pulp installation. See: https://pastebin.com/raw/S0U6k15y

PR #416, opened 2017-08-07
Add roles to populate, backup, and restore pulp

kdelee:
Adds two playbooks and three roles.

The `pulp_backup.yaml` playbook uses existing roles to install pulp on a
host, and then uses the new roles pulp-populate and pulp-backup to
populate it with content and back it up. Optionally, the user can fetch
the backup files to the control node.

The `pulp_restore.yaml` playbook uses existing roles to install pulp on
a host and uses the new role pulp-restore-backup to restore the files
and (optionally) test whether or not the restore succeeded.

These changes should be merged before #417, because #417 depends on these changes.

There are jenkins jobs using this branch that I can link you to to demonstrate how it works.

### DIFF
--- a/ci/ansible/roles/pulp-backup/defaults/main.yaml
+++ b/ci/ansible/roles/pulp-backup/defaults/main.yaml
@@ -1,0 +1,17 @@
+---
+fetch_to_localhost: true
+remote_bkup_dir: "{{ ansible_user_dir }}/remote_backups"
+local_bkup_dir: "backups"
+files_to_backup:
+  etc_pulp:
+    src_dir: /etc/pulp
+    file: etc_pulp
+  etc_pki_pulp:
+    src_dir: /etc/pki/pulp
+    file: etc_pki_pulp
+  var_lib_pulp:
+    src_dir: /var/lib/pulp
+    file: var_lib_pulp
+db_to_bkup:
+  dir: mongodb_backup_dir
+  file: mongodb_backup
--- a/ci/ansible/roles/pulp-backup/handlers/main.yaml
+++ b/ci/ansible/roles/pulp-backup/handlers/main.yaml
@@ -1,0 +1,12 @@
+---
+
+- name: Start Pulp
+  service: "name={{ item }} state=started"
+  with_items:
+    - httpd
+    - pulp_celerybeat
+    - pulp_resource_manager
+    - pulp_streamer
+    - pulp_workers
+    - qpidd
+
--- a/ci/ansible/roles/pulp-backup/tasks/main.yaml
+++ b/ci/ansible/roles/pulp-backup/tasks/main.yaml
@@ -1,0 +1,57 @@
+# rhel6 pulp 2.11.z backup
+- name: Stop Pulp
+  service: "name={{ item }} state=stopped"
+  with_items:
+    - httpd
+    - pulp_celerybeat
+    - pulp_resource_manager
+    - pulp_streamer
+    - pulp_workers
+    - qpidd
+  notify: Start Pulp
+
+- name: Install dependencies
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=present"
+  with_items:
+    - tar
+    - bzip2
+    - bzip2-libs
+
+- name: Create backups folder
+  file: path={{ remote_bkup_dir }} state=directory
+
+- name: Create local backups folder
+  local_action: file path={{ local_bkup_dir }} state=directory
+  become: false
+  when: fetch_to_localhost
+
+- name: Dump mongodb database
+  command: "mongodump --db pulp_database --out {{ remote_bkup_dir }}/{{ db_to_bkup.dir }}"
+
+- name: Archive filesystem backups
+  command: "tar -cjf {{ remote_bkup_dir }}/{{ item.value.file }}.tar.bz2 {{ item.value.src_dir }}/"
+  with_dict: "{{ files_to_backup }}"
+
+- name: Archive database backups
+  archive:
+    path: "{{ remote_bkup_dir }}/{{ db_to_bkup.dir }}"
+    dest: "{{ remote_bkup_dir }}/{{ db_to_bkup.file }}.bz2"
+    format: bz2
+
+- name: Fetch filesystem backups
+  fetch:
+    src: "{{ remote_bkup_dir }}/{{ item.value.file }}.tar.bz2"
+    dest: "{{ local_bkup_dir }}/{{ item.value.file }}.tar.bz2"
+    flat: yes
+  with_dict: "{{ files_to_backup }}"
+
+- name: Fetch database backups
+  fetch:
+    src: "{{ remote_bkup_dir }}/{{ db_to_bkup.file }}.bz2"
+    dest: "{{ local_bkup_dir }}/{{ db_to_bkup.file }}.bz2"
+    flat: yes
+  when: fetch_to_localhost
+
+- name: Delete temp backup folder from server
+  file: path={{ remote_bkup_dir }} state=absent
+  when: fetch_to_localhost
--- a/ci/ansible/roles/pulp-populate/tasks/main.yaml
+++ b/ci/ansible/roles/pulp-populate/tasks/main.yaml
@@ -1,0 +1,7 @@
+---
+- name: Populate the server
+  shell: >
+    source "{{ ansible_user_dir }}/.venvs/pulp-migrate/bin/activate" &&
+    pytest --junit-xml "{{ test_results_dir }}/populate.test.report.xml" --pyargs pulp_migrate.populate
+  ignore_errors: yes
+  # We would rather find out about errors from the xml report and not halt all execution
--- a/ci/ansible/roles/pulp-restore-backup/defaults/main.yaml
+++ b/ci/ansible/roles/pulp-restore-backup/defaults/main.yaml
@@ -1,0 +1,19 @@
+---
+remote_bkup_dir: "/root/backups"
+local_bkup_dir: "backups"
+files_to_restore:
+  etc_pulp:
+    dest_dir: /etc/pulp
+    file: etc_pulp
+    archive: etc_pulp.tar.bz2
+  etc_pki_pulp:
+    dest_dir: /etc/pki/pulp
+    file: etc_pki_pulp
+    archive: etc_pki_pulp.tar.bz2
+  var_lib_pulp:
+    dest_dir: /var/lib/pulp
+    file: var_lib_pulp
+    archive: var_lib_pulp.tar.bz2
+db_to_restore:
+  archive: mongodb_backup.bz2
+  dir: mongodb_backup_dir
--- a/ci/ansible/roles/pulp-restore-backup/tasks/main.yaml
+++ b/ci/ansible/roles/pulp-restore-backup/tasks/main.yaml
@@ -1,0 +1,69 @@
+# rhel7 pulp 2.13.z restore
+- name: Ensure expected variables
+  assert:
+    that:
+      - remote_bkup_dir is defined
+      - local_bkup_dir is defined
+    msg: |
+      Location of remote backup directory and local directory for backup to be copied to
+      must be specified with varibales remote_bkup_dir and local_bkup_dir
+
+- name: Install dependencies
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=present"
+  with_items:
+    - tar
+    - bzip2
+    - bzip2-libs
+
+- name: Stop Pulp
+  service: "name={{ item }} state=stopped"
+  with_items:
+    - httpd
+    - pulp_celerybeat
+    - pulp_resource_manager
+    - pulp_streamer
+    - pulp_workers
+    - qpidd
+  notify: Start Pulp
+
+- name: Make sure backup folder exists
+  file: "path={{ remote_bkup_dir }} state=directory"
+
+- name: Copy filesystem backups to remote
+  copy:
+    src: "{{ local_bkup_dir }}/{{ item.value.archive }}"
+    dest: "{{ remote_bkup_dir }}"
+  with_dict: "{{ files_to_restore }}"
+
+- name: Unarchive filesystem backups
+  command: "tar vxjf {{ remote_bkup_dir }}/{{ item.value.archive }} --overwrite"
+  args:
+    chdir: "/"
+  with_dict: "{{ files_to_restore }}"
+
+- name: Reset SELinux labels
+  command: "restorecon -r {{ item.value.dest_dir }}"
+  with_dict: " {{ files_to_restore }}"
+
+- name: Drop pulp database
+  command: "mongo pulp_database --eval 'db.dropDatabase()'"
+
+# unarchive module fails probably because
+# it relies on the using the "--diff" option which fails when
+# unarchiving a directory full of files that don't yet exist
+# See: https://github.com/ansible/ansible-modules-core/issues/2936#issuecomment-189579092
+- name: Copy database backups to remote
+  copy:
+    src: "{{ local_bkup_dir }}/{{ db_to_restore.archive }}"
+    dest: "{{ remote_bkup_dir }}"
+
+- name: Unarchive database backups
+  command: "tar vxjf {{ remote_bkup_dir }}/{{ db_to_restore.archive }}"
+  args:
+    chdir: "{{ remote_bkup_dir }}"
+
+- name: Restore database from backup
+  command: "mongorestore {{ remote_bkup_dir }}/{{ db_to_restore.dir }}"
+
+- name: Run pulp-manage-db
+  command: "sudo -u apache pulp-manage-db"
--- a/ci/ansible/roles/pulp-test-populate/tasks/main.yaml
+++ b/ci/ansible/roles/pulp-test-populate/tasks/main.yaml
@@ -1,0 +1,7 @@
+---
+- name: Test that the server is populated
+  shell: >
+    source "{{ ansible_user_dir }}/.venvs/pulp-migrate/bin/activate" &&
+    pytest --junit-xml "{{ test_results_dir }}/restore.test.report.xml" --pyargs pulp_migrate.test_restore
+  ignore_errors: yes
+  # We would rather find out about errors from the xml report and not halt all execution
--- a/ci/ansible/roles/subscription-manager/tasks/main.yaml
+++ b/ci/ansible/roles/subscription-manager/tasks/main.yaml
@@ -22,6 +22,7 @@
     username: "{{ rhn_username }}"
     password: "{{ rhn_password }}"
     pool: "{{ rhn_pool }}"  # e.g. rhn_pool='^SKU Name$'
+  ignore_errors: yes
   when:
     - rhn_pool is defined
     - rhn_username is defined
--- a/ci/ansible/pulp_backup.yaml
+++ b/ci/ansible/pulp_backup.yaml
@@ -1,0 +1,61 @@
+---
+
+- hosts: all
+  vars:
+    skip_populate: false
+  roles:
+    - role: subscription-manager
+      when: ansible_distribution == 'RedHat'
+    - role: pulp
+    - role: lazy
+      when: pulp_version | version_compare('2.8', '>=')
+    - role: pulp-certs
+    - role: pulp-crane
+    - role: pulp-populate
+      when: not skip_populate
+    - role: pulp-test-populate
+    - role: pulp-backup
+
+# This playbook runs roles to install pulp on a host, then populate it with
+# content and back up that content. It is possible to skip population by
+# setting "skip_populate" to "true"
+#
+# Be advised that this play will attempt to create a directory on the control
+# node to fetch the backed up files. This behavior can be controlled by setting
+# variables on invocation. See details below.
+#
+# Example invocation:
+#
+# ansible-playbook -i ~/hosts -l rhel6 pulp_backup.yaml \
+#                   -e local_bkup_dir=${HOME}/backups \
+#                   -e remote_bkup_dir=/root/backups \
+#                   -e rhn_password=my_pass \
+#                   -e 'rhn_pool=my_pool' \
+#                   -e rhn_username=my_username \
+#                   -e pulp_build='stable' \
+#                   -e pulp_version=2.11
+#
+# This would install 2.11 on the rhel6 host in your hosts file, and
+# fetch the backups to a directory named "backups" in your home directory.
+#
+# Variables you might want to specify on invocation:
+#
+# fetch_to_localhost: [boolean] default: TRUE
+#   When true, a directory is created on the control node in which
+#   to fetch the backups. If you have troubles with this, you may
+#   elect to create the directory in advance and pass:
+#   "-e local_bkup_dir='path/to/your/dir'"
+#
+# local_bkup_dir: [absolute path on control node] default "local_backups"
+#   This will create a directory, by default "local_backups" in the
+#   directory where the playbook was invoked.
+#
+# remote_bkup_dir: [absolute path on pulp host] default:
+#           "{{ ansible_user_dir }}/remote_backups"
+#   This will create a directory on the remote host in which to place the
+#   archives before they are fetched. When the files are fetched to the
+#   control node, this directory is destroyed.
+#
+# skip_populate: [boolean] default: FALSE
+#   Set this to TRUE if you want to not populate the server with fixture
+#   data.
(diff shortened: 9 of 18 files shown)


ci/ansible/roles/pulp-backup/tasks/main.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+    - name: Create local backups folder
+      local_action: file path={{ local_bkup_dir }} state=directory mode=0777
+      when: fetch_to_localhost
+
+    - name: Dump mongodb database
+      command: "mongodump --db pulp_database --out {{ remote_bkup_dir }}/{{ db_to_bkup.mongodb.dir }}"
+
+    - name: Archive filesystem backups
+      command: "tar -cjf {{ remote_bkup_dir }}/{{ item.value.file }}.tar.bz2 {{ item.value.src_dir }}/"
```
kdelee: I ended up using tar directly here because, from what I could find, no way to pass these options to the archive module at the moment and the standard options it uses don't preserve symlinks (needed for the published content)

kdelee:
@Ichimonji10 
Should be working now... fixed up pulp-migrate, but having trouble with jenkins just because it is being funny (yum is timing out trying to install ansible before any of this happens)

ci/ansible/roles/pulp-backup/tasks/main.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - pulp_workers
+    - qpidd
+  notify: Restart services
+
+- block:
+    - name: Create backups folder
+      file: path={{ remote_bkup_dir }} state=directory
+
+    - name: Create local backups folder
+      local_action: file path={{ local_bkup_dir }} state=directory mode=0777
```
kdelee: The file mode set on this directory created on the control node is hacky and I'd like a better alternative.

It may be better to just trash this step of creating the folder on the control node and require that the caller provide it. The only requirement it must fulfil is existence and permission to write to it.

kdelee:
@Ichimonji10 

Find notes on how to use in the playbooks.

If you have a hosts file that looks like this:

```
[rhel6]
rhel6.example.com

[rhel6:vars]
pulp_version=2.11
pulp_build=stable
ansible_user=root


[rhel7]
rhel6.example.com

[rhel7:vars]
pulp_version=2.13
pulp_build=stable
ansible_user=root
```

And a directory called `${HOME}/backups`,

You should be able to run the pulp_backup.yaml playbook on the rhel6 machine and then the pulp_restore.yaml playbook on the rhel7 machine as shown below:

```
$ ansible-playbook -i ~/hosts -l rhel6 pulp_backup.yaml \
                   -e local_bkup_dir=${HOME}/backups \
                   -e remote_bkup_dir=/root/backups \
                   -e rhn_password=my_pass \
                   -e 'rhn_pool=my_pool' \
                   -e rhn_username=my_username

# wait a while
$ tree ${HOME}/backups
/home/yourname/backups/
                 ├── etc_pki_pulp.tar.bz2 # archive of /etc/pki/pulp/
                 ├── etc_pulp.tar.bz2     # archive of /etc/pulp/
                 ├── mongodb_backup.bz2   # archive of mongodump of pulp_database
                 └── var_lib_pulp.tar.bz2 # archive of /var/lib/pulp

$ ansible-playbook -i ~/hosts -l rhel7 pulp_restore.yaml \
                     -e local_bkup_dir=${HOME}/backups \
                     -e remote_bkup_dir=/root/backups \
                     -e rhn_password=my_pass \
                     -e 'rhn_pool=my_pool' \
                     -e rhn_username=my_username  \
                     -e test_results_dir='/root'
```

ci/ansible/roles/pulp-backup/defaults/main.yaml
```diff
@@ -0,0 +1,20 @@
+---
+# location of backup on remote
+fetch_to_localhost: true
+remote_bkup_dir: "{{ ansible_user_dir }}/remote_backups"
+local_bkup_dir/: "local_backups"
```
Ichimonji10: Is the trailing slash on this variable's name significant?

ci/ansible/roles/pulp-populate/tasks/main.yaml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Install packages for population task
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=latest"
+  with_items:
+    - python34
+    - python34-setuptools
+    - gcc
+
+- name: Ensure services are running
```
Ichimonji10: Nitpick: Can the same term ("started") be used in both the task description and body?

ci/ansible/roles/pulp-populate/tasks/main.yaml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Install packages for population task
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=latest"
+  with_items:
+    - python34
+    - python34-setuptools
+    - gcc
```
Ichimonji10: Can `git` be adde to this list? A base install of RHEL 6 doesn't include it.

ci/ansible/roles/pulp-populate/tasks/main.yaml
```diff
@@ -0,0 +1,28 @@
...
+- name: Create config directory
+  file: path="{{ ansible_user_dir }}/.config/pulp_smash" state=directory
+
+- name: Ensure a Pulp Smash configuration file is present
+  template:
+    src: settings.j2
+    dest: "{{ ansible_user_dir }}/.config/pulp_smash/settings.json"
+
+- name: Populate the server
+  script: populate.sh "{{ test_results_dir }}"
```
Ichimonji10: This script can be dropped if the following playbook is used:

```yaml
---
- name: Install packages for population task
  action: "{{ ansible_pkg_mgr }} name={{ item }} state=latest"
  with_items:
    - python34
    - python34-setuptools
    - gcc
    - git

- name: Ensure services are running
  service: "name={{ item }} state=started"
  with_items:
    - httpd
    - pulp_celerybeat
    - pulp_resource_manager
    - pulp_streamer
    - pulp_workers
    - qpidd

- name: Create a directory for virtualenvs
  file:
    path: "{{ ansible_user_dir }}/.venvs"
    state: directory

- name: Create a virtualenv
  command: python3 -m venv "{{ ansible_user_dir }}/.venvs/pulp-migrate"
  args:
    creates: "{{ ansible_user_dir }}/.venvs/pulp-migrate"

# We fetch a zip with https://, because something asks for a password when this
# is used: git+https://git@github.com/PulpQE/pulp-migrate.git#egg=pulp-migrate
- name: Install pytest and pulp-migrate into the virtualenv
  shell: >
    source "{{ ansible_user_dir }}/.venvs/pulp-migrate/bin/activate" &&
    pip install pytest https://github.com/PulpQE/pulp-migrate/archive/master.zip#egg=pulp-migrate

- name: Create directory for pulp smash configuration file
  file:
    path: "{{ ansible_user_dir }}/.config/pulp_smash"
    state: directory

- name: Create pulp smash configuration file
  template:
    src: settings.j2
    dest: "{{ ansible_user_dir }}/.config/pulp_smash/settings.json"

- name: Populate the server
  shell: >
    source "{{ ansible_user_dir }}/.venvs/pulp-migrate/bin/activate" &&
    pytest --junit-xml "{{ test_results_dir }}/populate.test.report.xml" --pyargs pulp_migrate.populate
```

ci/ansible/roles/pulp-backup/handlers/main.yaml
```diff
@@ -0,0 +1,12 @@
+---
+
+- name: Restart services
```
Ichimonji10: This doesn't restart any services. It ensures services are started. The former always produces a change of state, and the latter only produces a change of state if needed. How about calling this "Start Pulp"?

ci/ansible/roles/pulp-backup/tasks/main.yaml
```diff
@@ -0,0 +1,52 @@
...
+  with_items:
+    - httpd
+    - pulp_celerybeat
+    - pulp_resource_manager
+    - pulp_streamer
+    - pulp_workers
+    - qpidd
+  notify: Restart services
+
+- block:
```
Ichimonji10: Why does this block declaration exist?

ci/ansible/roles/pulp-restore-backup/tasks/main.yaml
```diff
@@ -0,0 +1,99 @@
...
+  service: "name={{ item }} state=stopped"
+  with_items:
+    - httpd
+    - pulp_celerybeat
+    - pulp_resource_manager
+    - pulp_streamer
+    - pulp_workers
+    - qpidd
+
+- block:
```
Ichimonji10: Why does this block declaration exist?

ci/ansible/roles/pulp-restore-backup/tasks/main.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - name: Restore database from backup
+      command: "mongorestore {{ remote_bkup_dir }}/{{ db_to_restore.mongodb.dir }}"
+
+    - name: Run pulp-manage-db
+      command: "sudo -u apache pulp-manage-db"
+
+    - name: Delete temp backup folder from server
+      file: path={{ remote_bkup_dir }} state=absent
+
+- name: Start services
```
Ichimonji10: Maybe "Start Pulp"? More descriptive.

ci/ansible/roles/pulp-restore-backup/tasks/main.yaml
```diff
@@ -0,0 +1,99 @@
...
+  - name: Create config directory
+    file: path="{{ ansible_user_dir }}/.config/pulp_smash" state=directory
+
+  - name: Ensure a Pulp Smash configuration file is present
+    template:
+      src: settings.j2
+      dest: "{{ ansible_user_dir }}/.config/pulp_smash/settings.json"
+
+  - name: Test that content is restored
+    script: test_restore.sh "{{ test_results_dir }}"
```
Ichimonji10: I'm guessing that this script can be dropped, much like the populate script can be dropped.

ci/ansible/roles/pulp-backup/defaults/main.yaml
kdelee: I'm assuming that the directory name supplied does not have a trailing slash, so it is needed to form the path

ci/ansible/roles/pulp-backup/tasks/main.yaml
kdelee: I think it is an artifact of when I originally had this as a block you could skip in the back up step. I'll remove it

kdelee:
I pushed a commit with most of the changes you recommended. Actually pushing to this branch is the easiest way for me to run tests with jenkins and other remotes. I'll squash it back down to one commit once I'm sure this all works.

I did identify one issue that jenkins was running into when running the restore script locally, which I had yet to run into because I had always been using my machine as the control node against some remote rhel7 machine.

Still looking into if there is a work around to the subscription-manager problem I've been having.

ci/ansible/roles/pulp-backup/defaults/main.yaml
Ichimonji10: Are you sure? I believe every usage of `local_bkup_dir` already inserts slashes as appropriate:

```console
$ git grep local_bkup_dir
pulp_backup.yaml:#                   -e local_bkup_dir=${HOME}/backups \
roles/pulp-backup/tasks/main.yaml:    dest: "{{ local_bkup_dir }}/{{ item.value.file }}.tar.bz2"
roles/pulp-backup/tasks/main.yaml:    dest: "{{ local_bkup_dir }}/{{ item.value.file }}.bz2"
roles/pulp-restore-backup/tasks/main.yaml:    src: "{{ local_bkup_dir }}/{{ item.value.archive }}"
roles/pulp-restore-backup/tasks/main.yaml:    src: "{{ local_bkup_dir }}/{{ item.value.archive }}"
[snip!]
```

I also ran this playbook locally without the slash and didn't encounter any issues. Then again, I made some other changes while fiddling around, too.

ci/ansible/roles/pulp-test-populate/tasks/main.yaml
```diff
@@ -0,0 +1,52 @@
+---
+- name: Start Pulp
```
Ichimonji10: This role seems to mostly be identical to the `pulp-populate` role. This duplication can be eliminated by making a third role, and then having this and `pulp-populate` role both depend on it. Dependencies can be declared by creating a `{role_name}/meta/main.yml` file:

```yaml
---
dependencies:
  - { role: common, some_parameter: 3 }
  - { role: apache, apache_port: 80 }
  - { role: postgres, dbname: blarg, other_parameter: 12 }
```

kdelee: Ok, I'll try this out.

Can you share any insight you have on how variables are handled when using this method?

i.e. if options are passed `-e myvar='foo'` this is available at the global scope, right? 

I.e. if tasks are run because one role declares dependency on another, then the variable defined on the command line is available to all tasks, right?

ci/ansible/roles/pulp-backup/defaults/main.yaml
kdelee: Oh I was looking at the wrong line. That is absolutely a typo. Good eye. In fact, I imagine it sets a default variable to some value that is never used ( I imagine "{{ local_bkup_dir }}"  != "{{ local_bkup_dir/ }}"

ci/ansible/roles/pulp-test-populate/tasks/main.yaml
Ichimonji10: Variables passed at the command line with `-e` always, always, always win. They're the highest priority type of variable, and will override whatever's in a playbook.

ci/ansible/roles/subscription-manager/tasks/main.yaml
```diff
@@ -22,6 +22,7 @@
     username: "{{ rhn_username }}"
     password: "{{ rhn_password }}"
     pool: "{{ rhn_pool }}"  # e.g. rhn_pool='^SKU Name$'
+  ignore_errors: yes
```
kdelee: I think this is safe to do for all rhel jobs because if the subscription _really_ did fail, then enabling the repos would fail and the job would abort.

kdelee:
@Ichimonji10 
I have squashed this down to one commit and had a successful run of the two jenkins jobs using it.
Running it once more now.

This morning I went down a rabbit hole trying to use a different SKU this morning with John from candlepin and Preethi, but alas, a can a of worms. I think it is best addressed in its own PR and some other work outside the scope of this repository.

 Right now ignoring the errors on the subscribe step seems to be the best since it throws return code of 70 even on success 😞

ci/ansible/roles/pulp-backup/tasks/main.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - bzip2
+    - bzip2-libs
+
+- name: Create backups folder
+  file: path={{ remote_bkup_dir }} state=directory
+
+- name: Create local backups folder
+  local_action: file path={{ local_bkup_dir }} state=directory
+  become: false
+  when: fetch_to_localhost
```
kdelee: Got rid of the hacky file mode that I had here and use "become: false" so that the user that executed the ansible-playbook is used.

kdelee:
@Ichimonji10 I have run this several times now and it seems stable.

Can I get a final review? Should I ask for other reviewers?

Ichimonji10:
I'm reviewing this now.

Ichimonji10:
If this diff is applied, then the `pulp-migrate-tools` role is fully idempotent:

```diff
diff --git a/ci/ansible/roles/pulp-migrate-tools/tasks/main.yaml b/ci/ansible/roles/pulp-migrate-tools/tasks/main.yaml
index abf1888..b97ba27 100644
--- a/ci/ansible/roles/pulp-migrate-tools/tasks/main.yaml
+++ b/ci/ansible/roles/pulp-migrate-tools/tasks/main.yaml
@@ -26,6 +26,7 @@
   command: python3 -m venv "{{ ansible_user_dir }}/.venvs/pulp-migrate"
   args:
     creates: "{{ ansible_user_dir }}/.venvs/pulp-migrate"
+  register: result
 
 # We fetch a zip with https://, because something asks for a password when this
 # is used: git+https://git@github.com/PulpQE/pulp-migrate.git#egg=pulp-migrate
@@ -33,6 +34,7 @@
   shell: >
     source "{{ ansible_user_dir }}/.venvs/pulp-migrate/bin/activate" &&
     pip install pytest https://github.com/PulpQE/pulp-migrate/archive/master.zip#egg=pulp-migrate
+  when: result|changed
 
 - name: Create directory for pulp smash configuration file
   file:
```

ci/ansible/roles/pulp-backup/tasks/main.yaml
```diff
@@ -0,0 +1,59 @@
+# rhel6 pulp 2.11.z backup
+- name: Stop services before backup
```
Ichimonji10: Nitpick: `s/services/Pulp/`

ci/ansible/roles/pulp-backup/defaults/main.yaml
```diff
@@ -0,0 +1,20 @@
...
+    file: etc_pulp
+  etc_pki_pulp:
+    src_dir: /etc/pki/pulp
+    file: etc_pki_pulp
+  var_lib_pulp:
+    src_dir: /var/lib/pulp
+    file: var_lib_pulp
+
+db_to_bkup:
+  mongodb:
```
Ichimonji10: Is this extra level of indirection necessary? I think that this variable could be written as:

```yaml
db_to_bkup
  dir: mongodb_backup_dir
  file: mongodb_backup
```

The references to this variable could then be simplified. Instead of this:

```yaml
- name: Archive database backups
  archive:
    path: "{{ remote_bkup_dir }}/{{ item.value.dir }}"
    dest: "{{ remote_bkup_dir }}/{{ item.value.file }}.bz2"
    format: bz2
  with_dict: "{{ db_to_bkup }}"
```

One could have this:

```yaml
- name: Archive database backup
  archive:
    path: "{{ remote_bkup_dir }}/{{ db_to_bkup.dir }}"
    dest: "{{ remote_bkup_dir }}/{{ db_to_bkup.file }}"
    format: bz2
```

Ichimonji10:
```diff
diff --git a/ci/ansible/roles/pulp-backup/defaults/main.yaml b/ci/ansible/roles/pulp-backup/defaults/main.yaml
index 5f8dc52..bab29cb 100644
--- a/ci/ansible/roles/pulp-backup/defaults/main.yaml
+++ b/ci/ansible/roles/pulp-backup/defaults/main.yaml
@@ -15,6 +15,5 @@ files_to_backup:
     file: var_lib_pulp
 
 db_to_bkup:
-  mongodb:
-    dir: mongodb_backup_dir
-    file: mongodb_backup
+  dir: mongodb_backup_dir
+  file: mongodb_backup
diff --git a/ci/ansible/roles/pulp-backup/tasks/main.yaml b/ci/ansible/roles/pulp-backup/tasks/main.yaml
index 575354d..9d60023 100644
--- a/ci/ansible/roles/pulp-backup/tasks/main.yaml
+++ b/ci/ansible/roles/pulp-backup/tasks/main.yaml
@@ -26,7 +26,7 @@
   when: fetch_to_localhost
 
 - name: Dump mongodb database
-  command: "mongodump --db pulp_database --out {{ remote_bkup_dir }}/{{ db_to_bkup.mongodb.dir }}"
+  command: "mongodump --db pulp_database --out {{ remote_bkup_dir }}/{{ db_to_bkup.dir }}"
 
 - name: Archive filesystem backups
   command: "tar -cjf {{ remote_bkup_dir }}/{{ item.value.file }}.tar.bz2 {{ item.value.src_dir }}/"
@@ -34,10 +34,9 @@
 
 - name: Archive database backups
   archive:
-    path: "{{ remote_bkup_dir }}/{{ item.value.dir }}"
-    dest: "{{ remote_bkup_dir }}/{{ item.value.file }}.bz2"
+    path: "{{ remote_bkup_dir }}/{{ db_to_bkup.dir }}"
+    dest: "{{ remote_bkup_dir }}/{{ db_to_bkup.file }}.bz2"
     format: bz2
-  with_dict: "{{ db_to_bkup }}"
 
 - name: Fetch filesystem backups
   fetch:
@@ -48,10 +47,9 @@
 
 - name: Fetch database backups
   fetch:
-    src: "{{ remote_bkup_dir }}/{{ item.value.file }}.bz2"
-    dest: "{{ local_bkup_dir }}/{{ item.value.file }}.bz2"
+    src: "{{ remote_bkup_dir }}/{{ db_to_bkup.file }}.bz2"
+    dest: "{{ local_bkup_dir }}/{{ db_to_bkup.file }}.bz2"
     flat: yes
-  with_dict: "{{ db_to_bkup }}"
   when: fetch_to_localhost
 
 - name: Delete temp backup folder from server
```

ci/ansible/roles/pulp-backup/defaults/main.yaml
```diff
@@ -0,0 +1,20 @@
+---
+# location of backup on remote
```
Ichimonji10: Is this comment out of place?

ci/ansible/roles/pulp-restore-backup/defaults/main.yaml
```diff
@@ -0,0 +1,21 @@
...
+  etc_pki_pulp:
+    dest_dir: /etc/pki/pulp
+    file: etc_pki_pulp
+    archive: etc_pki_pulp.tar.bz2
+  var_lib_pulp:
+    dest_dir: /var/lib/pulp
+    file: var_lib_pulp
+    archive: var_lib_pulp.tar.bz2
+db_to_restore:
+  mongodb:
```
Ichimonji10: Same question here as for `db_to_bkup`. Is the extra level of indirection necessary? Or could this variable be written as:

```yaml
db_to_restore:
  archive: mongodb_backup.bz2
  dir: mongodb_backup_dir
```

ci/ansible/roles/pulp-restore-backup/tasks/main.yaml
```diff
@@ -0,0 +1,71 @@
...
+      must be specified with varibales remote_bkup_dir and local_bkup_dir
+
+- name: Install dependencies
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=present"
+  with_items:
+    - tar
+    - bzip2
+    - bzip2-libs
+
+- name: Stop services
```
Ichimonji10: Nitpick: `s/services/Pulp/`

ci/ansible/roles/pulp-restore-backup/tasks/main.yaml
```diff
@@ -0,0 +1,71 @@
+# rhel6 pulp 2.11.z backup
```
Ichimonji10: Is this comment out of place?

Ichimonji10:
This looks nice. I'm running the `pulp_backup.yaml` and `pulp_restore.yaml` playbooks right now.

ci/ansible/roles/pulp-backup/defaults/main.yaml
```diff
@@ -0,0 +1,20 @@
...
+fetch_to_localhost: true
+remote_bkup_dir: "{{ ansible_user_dir }}/remote_backups"
+local_bkup_dir: "local_backups"
```
Ichimonji10: The `pulp-backup` and `pulp-restore` roles have different default values for the `local_bkup_dir` variable. This needs to be fixed.

Ichimonji10:
You can see a summary of the changes I've suggested at [my repo](https://github.com/Ichimonji10/pulp_packaging/tree/pr-416). I'm spinning up a fresh pair of VMs and testing them now.

ci/ansible/roles/pulp-backup/defaults/main.yaml
kdelee: Yeah that is reasonable. At one point I had it like this but then changed it to be more like the files "in case" there was some other database we would want to back up. But that is probably never going to happen so I agree this is a good change.

ci/ansible/roles/pulp-backup/defaults/main.yaml
kdelee: yes

kdelee:
@Ichimonji10 I tested these changes locally and they work for me too
I'll let jenkins run them and then sqaush your changes in, then you think we are done here?

Ichimonji10:
Yup. That done, I think we're done with this PR.

kdelee:
Jenkins likes it. Thanks a lot for your help!

I've squashed your changes into the commit and I think it is good to go

PR #417, opened 2017-08-07
Add jobs to backup and restore pulp

kdelee:
Add pulp-restore and pulp-backup job templates that generate jobs to
test that pulp can be backed up and the data can be migrated to a
different version and OS.

These changes don't make sense to merge until #416 is merged.

I can link you to jobs that use the #416 branch upon request!

### DIFF
--- a/ci/jjb/jobs/pulp-backup.yaml
+++ b/ci/jjb/jobs/pulp-backup.yaml
@@ -1,0 +1,50 @@
+- job-template:
+    name: 'pulp-{pulp_version}-{pulp_build}-backup-{os}'
+    node: '{os}-np'
+    properties:
+        - qe-ownership
+        # don't want to build up too many artifacts because they are of
+        # some appreciable size
+        - build-discarder:
+            days-to-keep: 7
+            num-to-keep: 7
+            artifact-days-to-keep: 7
+            artifact-num-to-keep: 3
+    scm:
+        - pulp-packaging-github
+    wrappers:
+        - config-file-provider:
+            files:
+                - file-id: rhn_credentials
+                  variable: RHN_CREDENTIALS
+        - inject:
+            properties-content: |
+                OS={os}
+                PULP_BUILD={pulp_build}
+                PULP_VERSION={pulp_version}
+    triggers:
+        - timed: '@midnight'
+    builders:
+        - shell:
+            !include-raw-escape:
+                - 'ssh-setup.sh'
+        - shell:
+            !include-raw-escape:
+                - 'ansible-setup.sh'
+                - 'pulp-backup.sh'
+        - capture-logs
+    publishers:
+      - postbuildscript:
+          script-only-if-succeeded: False
+          script-only-if-failed: False
+          builders:
+            - shell: |
+                if [[ "${{OS}}" =~ "rhel" ]]; then
+                    sudo subscription-manager unregister
+                fi
+      - archive:
+          artifacts: '*.bz2'
+      - junit:
+          results: "*.xml"
+      - email-notify-owners
+      - mark-node-offline
--- a/ci/jjb/jobs/pulp-restore.yaml
+++ b/ci/jjb/jobs/pulp-restore.yaml
@@ -1,0 +1,48 @@
+- job-template:
+    name: 'pulp-{pulp_version}-{pulp_build}-restore-{os}'
+    node: '{os}-np'
+    properties:
+        - qe-ownership
+    scm:
+        - pulp-packaging-github
+    wrappers:
+        - config-file-provider:
+            files:
+                - file-id: rhn_credentials
+                  variable: RHN_CREDENTIALS
+        - inject:
+            properties-content: |
+                OS={os}
+                PULP_BUILD={pulp_build}
+                PULP_VERSION={pulp_version}
+    triggers:
+        - reverse:
+            jobs: 'pulp-{backup_version}-{backup_build}-backup-{backup_os}'
+            result: 'success'
+    builders:
+        - copyartifact:
+            project: 'pulp-{backup_version}-{backup_build}-backup-{backup_os}'
+            which-build: last-successful
+            filter: "*.bz2"
+            flatten: true
+        - shell:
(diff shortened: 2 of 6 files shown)


ci/jjb/jobs/pulp-backup.yaml
```diff
@@ -0,0 +1,47 @@
...
+                sudo subscription-manager unregister
+                sudo subscription-manager clean
+            fi
+        - capture-logs
+    publishers:
+      - archive:
+          artifacts: '*.bz2'
+      - join-trigger:
+          projects:
+            - 'pulp-2.13-stable-restore-rhel7'
```
elyezer: Since you are using a job template, I think it would be better to create a variable for the pulp version which the backup should be restored, something like restore_pulp_version so you can control on the permutation which should be the triggered job. For example you may have 2.14 version to restore a backup from 2.12 and in the current definition you can't control that.

ci/jjb/jobs/pulp-backup.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - junit:
+          results: "*.xml"
```
elyezer: Where the resulting XML file is being generated?

ci/jjb/jobs/pulp-backup.yaml
elyezer: I think this might not be necessary since you are using a reverse trigger on the restore job. What is the idea of having the join-trigger in here?

ci/jjb/jobs/pulp-restore.yaml
```diff
@@ -0,0 +1,43 @@
...
+                - file-id: rhn_credentials
+                  variable: RHN_CREDENTIALS
+        - inject:
+            properties-content: |
+                OS={os}
+                PULP_BUILD={pulp_build}
+                PULP_VERSION={pulp_version}
+    triggers:
+        - reverse:
+            jobs: 'pulp-2.11-stable-backup-rhel6'
```
elyezer: It would be better to control this reverse trigger by a JJB variable, you may want to restore from a backup on 2.12 for example.

ci/jjb/scripts/pulp-backup.sh
```diff
@@ -0,0 +1,14 @@
+sudo yum -y install ansible attr git libselinux-python
+echo 'localhost' > hosts
+source "${RHN_CREDENTIALS}"
+export ANSIBLE_CONFIG="${PWD}/ci/ansible/ansible.cfg"
```
elyezer: The lines 1-4 are being repeated, maybe we could extract them to a separated file and include it whenever we want to run an ansible playbook.

Also we should the hosts definition outside that file. The hosts, specially when using localhost, can be passed directly to the `ansible-playbook` command like `ansible-playbook [other options] -i "localhost,"`. The trailing comma is required otherwise ansible will think it is a file and not a list of hosts being passed directly.

ci/jjb/jobs/pulp-restore.yaml
```diff
@@ -0,0 +1,43 @@
...
+                - 'pulp-restore.sh'
+        - shell: |
+            if [[ "${{OS}}" =~ "rhel" ]]; then
+                sudo subscription-manager unregister
+                sudo subscription-manager clean
+            fi
+        - capture-logs
+    publishers:
+        - junit:
+            results: "*.xml"
```
elyezer: Where the resulting XML file is being generated?

ci/jjb/jobs/pulp-backup.yaml
kdelee: I was following the example I saw on other jobs of both having the upstream job trigger the downstream job and the downstream job having the reverse trigger. It can be removed if you think that is best.

ci/jjb/jobs/pulp-backup.yaml
kdelee: This is being generated during a task run in the ansible playbook

ci/jjb/jobs/pulp-restore.yaml
kdelee: If I add a new variable, will I have to put it in the job name? I ran into an issue where the pulp_build would not get unpacked out of list unless it was in the job name. So it was resolving as `['stable']` instead of `'stable'`

ci/jjb/jobs/pulp-restore.yaml
kdelee: This is being generated during a task run in the ansible playbook.

ci/jjb/scripts/pulp-backup.sh
kdelee: Ok I can try this out.

ci/jjb/jobs/pulp-restore.yaml
elyezer: @kdelee you don't need to put on the name. For example:

https://github.com/pulp/pulp_packaging/blob/master/ci/jjb/jobs/projects.yaml#L61 and https://github.com/pulp/pulp_packaging/blob/master/ci/jjb/jobs/pulp-dev.yaml#L20

ci/jjb/jobs/pulp-backup.yaml
kdelee: Ok, well I think I had fused examples from a couple different jobs. I'm going to try it with just the reverse trigger from the rhel7 job

kdelee:
Ok I've made the changes and am testing them out. 

Realized the rhel7 images we were getting did not have bzip2 libs installed and that was breaking these jobs, so hopefully I can get in a few successful test runs tomorrow to make sure everything is as it should be.

kdelee:
@elyezer 

I realized that if the build fails after subscription, but before the unsubscription, that we may have been eating up entitlements associated with the account we've been using. I've moved the unsubscribe step to a post-build action to run no matter if the build succeeds or fails.

Would you want me to update other jobs to do this as well on this PR or would that be a separate PR?

Other than that I think these are ready.

Ichimonji10:
:tada:

PR #418, opened 2017-08-08
Reformat Ansible play

Ichimonji10:
Ensure the following rules are followed:

* Use two spaces for each level of indentation.
* Use YAML syntax when passing arguments to Ansible modules, instead of
  Ansible's inline syntax.
* Use single quotes instead of double quotes where possible. (Yaml
  requires escaping backslashes in double quotes but not in single
  quotes.)
* Don't use back-slashes to wrap lines unless necessary.

### DIFF
--- a/ci/ansible/roles/pulp-certs/tasks/main.yml
+++ b/ci/ansible/roles/pulp-certs/tasks/main.yml
@@ -5,27 +5,32 @@
     name: openssl-perl
     state: present
   when:
-    - ansible_distribution == "Fedora"
+    - ansible_distribution == 'Fedora'
     - ansible_distribution_major_version|int == 26
 
 - name: Create CA to sign all certificates
   command: >
     /usr/bin/openssl req -x509 -newkey rsa:2048 -keyout private/cakey.pem -nodes -days 3650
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against F24 and F25 with Pulp 2.14 beta 3. Testing against RHEL 7 with Pulp 2.14 beta 3.

ci/ansible/roles/pulp-certs/tasks/main.yml
```diff
@@ -5,42 +5,48 @@
...
 - name: Update CA trust
   shell: update-ca-trust
 
 - name: Configure openssl subjectAltName
-  lineinfile: dest=/etc/pki/tls/openssl.cnf
-              insertafter="^\[ usr_cert \]"
-              line="subjectAltName=DNS:{{ ansible_fqdn }},DNS:{{ ansible_nodename }},DNS:{{ ansible_hostname }}"
+  lineinfile:
+    dest: /etc/pki/tls/openssl.cnf
+    insertafter: '^\[ usr_cert \]'
```
Ichimonji10: I recently found out that yaml treats single and double quotes differently. This produces a syntax error:

```
insertafter: "^\[ usr_cert \]"
```

PR #430, opened 2017-09-25
Added logging for case where redmine issue linked as external by bugz…

dralley:
…illa does not exist

### DIFF
--- a/ci/redmine_bugzilla.py
+++ b/ci/redmine_bugzilla.py
@@ -43,7 +43,7 @@ import ConfigParser
 import os
 
 from bugzilla.rhbugzilla import RHBugzilla
-from redminelib import Redmine
+from redminelib import Redmine, exceptions
 import xmlrpclib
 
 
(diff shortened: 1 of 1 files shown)


ci/redmine_bugzilla.py
```diff
@@ -250,7 +250,10 @@ def main():
                 if external_bug['type']['description'] == 'Pulp Redmine':
                     add_cc_list_to_bugzilla_bug(bug)
                     issue_id = external_bug['ext_bz_bug_id']
-                    issue = redmine.issue.get(issue_id)
+                    try:
+                        issue = redmine.issue.get(issue_id)
+                    except exceptions.ResourceNotFoundError:
+                        print "Issue ID {} linked as external on BZ {} does not exist".format(issue_id, bug.id)
```
mhrivnak: I think this block should re-raise. Otherwise you'll get an exception a couple lines later about reference before assignment.

dralley: Resolved

PR #438, opened 2017-10-12
Add JJB job to install Pulp 3

Ichimonji10:
Add a new job called "pulp-3-installer," which installs Pulp 3 on an
arbitrary host, over SSH, using Ansible. This new installer is very
similar to the "pulp-installer" job. From the job description:

> This job requires two hosts: one for use as an Ansible control host,
> and one on which to install Pulp 3. Jenkins is responsible for
> providing the control host, and you are responsible for providing the
> Pulp 3 host. When this job is executed, Jenkins will configure the
> control host, by doing things like installing Ansible and configuring
> SSH keys. Then, the control host will reach out and install Pulp 3
> onto the host identified by `HOSTNAME`.
>
> [snip!]
>
> Currently, this job can't install Pulp 3 on a RHEL host. (It can
> install Pulp 3 on a Fedora host, though.) This is because neither this
> job nor the Ansible installer it calls can deal with
> subscription-manager.

### DIFF
--- a/ci/jjb/jobs/pulp-3-installer.yaml
+++ b/ci/jjb/jobs/pulp-3-installer.yaml
@@ -1,0 +1,79 @@
+# Install Pulp 3 on a host.
+- job:
+    name: pulp-3-installer
+    concurrent: true
+    node: 'fedora-np'
+    description: |
+      <p>
+        Install Pulp on the host identified by the job parameter
+        <code>HOSTNAME</code>.
+      </p>
+      <p>
+        This job requires two hosts: one for use as an Ansible control host, and
+        one on which to install Pulp 3. Jenkins is responsible for providing the
+        control host, and you are responsible for providing the Pulp 3 host.
+        When this job is executed, Jenkins will configure the control host, by
+        doing things like installing Ansible and configuring SSH keys. Then, the
+        control host will reach out and install Pulp 3 onto the host identified
+        by <code>HOSTNAME</code>.
+      </p>
+      <p>
+        Ensure the following is in <code>~/.ssh/authorized_keys</code> on the
+        Pulp 3 host:
+      </p>
+      <pre>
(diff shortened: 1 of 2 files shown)


Ichimonji10:
I've tested this change out on my local system as best as I can. I don't have things like nodepool, though, so it'll probably require after-the-fact adjustments.

@preethit I think that you might find this job helpful.

Ichimonji10:
Note that this job is only capable of acting on Fedora hosts. It can't easily act on RHEL hosts, as the Ansible playbooks it calls are incapable of managing subscription information. (And updating those playbooks is prohibitively hard at this time. And I'd rather not do so, as doing so would increase our technical debt in that area.)

ci/jjb/jobs/pulp-3-installer.yaml
```diff
@@ -0,0 +1,51 @@
...
+    builders:
+      - shell: |
+          sudo yum install -y ansible python3
+          cd ansible
+          ansible-galaxy install -r requirements.yml -p roles
+          export ANSIBLE_HOST_KEY_CHECKING=False
+          ansible all -i "$HOSTNAME", -m selinux -a state=permissive
+          ansible-playbook deploy-pulp3.yml \
+            --inventory "$HOSTNAME", \
+            --private-key ../pulp_server_key
```
Ichimonji10: I have no idea where this file comes from. It's magical. Perhaps the [jenkins-ssh-credentials](https://github.com/pulp/pulp_packaging/blob/master/ci/jjb/jobs/macros.yaml#L213-L226) macro creates it?

The existing [pulp-installer](https://github.com/pulp/pulp_packaging/blob/master/ci/jjb/jobs/pulp-installer.yaml#L60) job also references this file, which is why I'm referencing it here.

ci/jjb/jobs/pulp-3-installer.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - jenkins-ssh-credentials
+    builders:
+      - shell: |
+          sudo yum install -y ansible python3
+          cd ansible
+          ansible-galaxy install -r requirements.yml -p roles
+          export ANSIBLE_HOST_KEY_CHECKING=False
+          ansible all \
+            --inventory "$HOSTNAME", \
+            --private-key ../pulp_server_key \
```
Ichimonji10: I have no idea where this file comes from. It's magical. Perhaps the [jenkins-ssh-credentials](https://github.com/pulp/pulp_packaging/blob/master/ci/jjb/jobs/macros.yaml#L213-L226) macro creates it? The existing [pulp-installer](https://github.com/pulp/pulp_packaging/blob/master/ci/jjb/jobs/pulp-installer.yaml#L60) job also references this file, which is why I'm referencing it here.

Ichimonji10:
This needs to be tested against our official Jenkins instance.

Ichimonji10:
Tested against our instance of Jenkins and a beaker box.

Ichimonji10:
Rebased on upstream/master.

PR #439, opened 2017-10-12
Add Jenkins job to install and smash Pulp 3

Ichimonji10:
The new job is called `pulp-3-dev-{os}`. It can be run against Fedora 25
and 26. It can't be run against RHEL, as neither it nor the Ansible
playbooks it calls have logic for working with subscription manager.

### DIFF
--- a/ci/jjb/jobs/projects.yaml
+++ b/ci/jjb/jobs/projects.yaml
@@ -70,6 +70,14 @@
         - pulp-{pulp_version}-dev-{os}
 
 - project:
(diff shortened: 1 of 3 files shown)


Ichimonji10:
Tested on my local Jenkins instance against F25 and F26 slaves.

Ichimonji10:
I've updated this PR so that it correctly disables SELinux.

Ichimonji10:
This PR has been tested against our official Jenkins instance, and it works.

PR #443, opened 2017-10-19
Add jobs to smash koji pulp nightly repo

pcreech:
Add jobs to smash koji pulp nightly repo

### DIFF
--- a/ci/ansible/roles/pulp/tasks/main.yaml
+++ b/ci/ansible/roles/pulp/tasks/main.yaml
@@ -97,6 +97,17 @@
       gpgcheck: 0
   when: pulp_build == "nightly"
 
+- name: Setup Pulp nightly koji repository
+  template:
+    src: yum_repo.j2
+    dest: /etc/yum.repos.d/{{ item.key }}.repo
+  with_dict:
+    pulp:
+      name: Pulp Project repository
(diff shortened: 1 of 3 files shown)


ci/ansible/roles/pulp/tasks/main.yaml
```diff
@@ -97,6 +97,17 @@
...
 
+- name: Setup Pulp nightly koji repository
+  template:
+    src: yum_repo.j2
+    dest: /etc/yum.repos.d/{{ item.key }}.repo
+  with_dict:
+    pulp:
+      name: Pulp Project repository
+      baseurl: "http://koji.katello.org/releases/yum/pulp-nightly/pulp/{% if ansible_distribution == 'Fedora' %}fedora-{% endif %}$releasever/$basearch/"
+      gpgcheck: 0
```
ehelms: I'm curious why you don't use the `yum_repository` module here? Is it an older version of Ansible being used?

pcreech: I copy-pasted it from the other repo definitions in the file and modified it for koji's location.  Looks like ansible-2.4.0 is installed on the machines running the jobs, so conversion to yum_repository might be worth evaluating.

PR #444, opened 2017-10-19
Make pulp-3-dev-{os} jobs smash Pulp 3

Ichimonji10:
Of note is that the pulp-smash-runner job is not being called. Instead,
a few extra shell commands are being executed.

This approach definitely has some downsides. Among other things, the
existing pulp-smash-runner job is highly configurable, where a user can
set options such as the broker used by the Pulp installation under test.
(e.g. qpid or rabbitmq) However, the job is also very complicated,
making calls to it much harder to accomplish. And the job performs some
tasks that are specific to Pulp 2, such as working with Pulp 2's SSL
certificates.

In this case, it seems better to have a good job now rather than a
perfect job later. This is doubly true if Jenkins 2 will soon be
available, which will make things like pipelines available.

### DIFF
--- a/ci/jjb/jobs/pulp-3-dev.yaml
+++ b/ci/jjb/jobs/pulp-3-dev.yaml
@@ -11,10 +11,6 @@
         installed on localhost. As a result, actions like SSH configuration are
         skipped.
       </p>
-      <p>
-        At this time, the "test" consists of running a simple command to verify
-        that Pulp Smash is correctly installed.
-      </p>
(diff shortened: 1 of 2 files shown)


Ichimonji10:
I tested this PR by creating and executing `jaudet-pulp-3-dev-f25` and `jaudet-pulp-3-dev-f26` on our Jenkins instance. The runs succeeded, and Jenkins is correctly reading the XML reports.

PR #445, opened 2017-10-30
Specify a port number in Pulp Smash's config file

Ichimonji10:
See: https://github.com/PulpQE/pulp-smash/pull/802


Ichimonji10:
This must be merged at the same time as the PR mentioned in the commit message in order to avoid breakage.

PR #447, opened 2017-10-31
Modify location for pulp-packaging

pcreech:
pulp-packaging is being moved to the pulp organization in github


pcreech:
Yeah, i ran across that one grep'ing for pulp-packaging in the codebase.  I've identified a few items that could use modification/update after today.

Ichimonji10:
I think https://github.com/pulp/pulp-ci/pull/448 addresses one of those renames.

PR #450, opened 2017-11-01
Clean up old jjb builds

werwty:
build-automation, docs-build-nightly, docs-build-manually, pulp-dev
has the 2.12-dev and 2.13-dev pulp versions removed, since no changes
are being made to those releases.

pulp-packaging-build has unnecessary excludes removed

pulp-upgrade matrix has been updated to only try to upgrade to the
latest pulp.

sync has been removed. This job was used to sync triage logs. The
new triage logs are syncing with a cronjob.

unittest jobs has old versions removed, since no PRs should be landing
against the current dev-branch.

### DIFF
--- a/ci/jjb/jobs/unittests.yaml
+++ b/ci/jjb/jobs/unittests.yaml
@@ -46,18 +46,6 @@
           allow-whitelist-orgs-as-admins: true
           # only-trigger-phrase: true
           white-list-target-branches:
-            - 2.8-dev
-            - 2.8-release
-            - 2.9-dev
-            - 2.9-release
-            - 2.10-dev
-            - 2.10-release
-            - 2.11-dev
-            - 2.11-release
-            - 2.12-dev
-            - 2.12-release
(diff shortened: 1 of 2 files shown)


werwty:
@preethit Could you also leave a review? For some reason I can't add you to Reviewers.

ci/jjb/jobs/unittests.yaml
```diff
@@ -10,6 +10,7 @@
           values:
             - f24-np
             - f25-np
+            - f26-np
```
Ichimonji10: AFAIK, Pulp master (2.15) doesn't successfully install on Fedora 26. As a result, using it for unit tests might cause lots of failures. You might want to do this addition separately from the clean-ups done elsewhere in this PR.

Ichimonji10: Ahh, I see. It makes sense that the unit tests wouldn't require a working Pulp installation. Thanks for the info. :smile:

PR #453, opened 2017-11-07
Fix pulp packaging jobs

pcreech:
Fix pulp packaging jobs

### DIFF
--- a/ci/jjb/jobs/projects.yaml
+++ b/ci/jjb/jobs/projects.yaml
@@ -21,11 +21,11 @@
     os:
         - 'f24'
         - 'f25'
+        - 'f26'
         - 'rhel7'
     pulp_version:
         - '2.15':
             reverse_trigger: 'master'
-    reverse_trigger: '{pulp_version}-dev'
     jobs:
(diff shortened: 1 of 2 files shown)


ci/jjb/jobs/projects.yaml
```diff
@@ -21,11 +21,11 @@
     os:
         - 'f24'
         - 'f25'
+        - 'f26'
         - 'rhel7'
     pulp_version:
-        - '2.15':
+        - 'master':
```
Ichimonji10: This change will break the coverage report functionality in `pulp-dev.yaml`. Search that file for `PULP_VERSION` and you'll see what I mean.

I don't think that the coverage report works that well right now, and I'm not aware of anyone that uses it. Still, it'd be good to either fix or drop that bit of code.

pcreech: Since this is for pulp-packaging, it should be ok (the pulp-dev job definition still has this setting set appropriately)

Ichimonji10: Sorry. I referenced the file. I think this change will break the coverage report defined in `pulp-packaging-build.yaml`.

PR #455, opened 2017-11-20
Separate pulp3-docs builder jobs from pulp2-docs builder jobs.

werwty:
Pulp3 has to be installed and running before rest api docs can be
built. We should install pulp3 from ansible roles and start pulp3
server in background with django runserver, before building the docs.

closes #2866
https://pulp.plan.io/issues/2866

### DIFF
--- a/ci/jjb/jobs/docs.yaml
+++ b/ci/jjb/jobs/docs.yaml
@@ -1,5 +1,5 @@
 # This set of Jenkins jobs which build the Pulp platform and plugin documentation and publishes
-# it to https://docs.pulpproject.org/
+# it to https://docs.pulpproject.org/ for pulp2
 
 - job-template:
     name: 'docs-builder-{release_config}'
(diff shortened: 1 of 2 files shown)


ci/jjb/jobs/docs.yaml
```diff
@@ -64,3 +64,114 @@
...
+            ansible all \
+                --inventory localhost, \
+                --connection local \
+                --become \
+                --become-user pulp \
+                -m shell \
+                -a 'source ~/pulpvenv/bin/activate && pulp-manager runserver localhost:8000 &'
+
+            # clone and build the docs
+            cd ../../pulp-ci/ci/
```
werwty: @Ichimonji10 Is there a better way to go back to the workspace?

Ichimonji10: `pushd` and `popd` let you manipulate a stack of directories.

```console
$ pwd
/home/ichimonji10
$ pushd ~/Downloads  # push a directory onto the stack
~/Downloads ~
$ cd ~/Documents  # change the top directory on the stack
$ dirs  # view the stack
~/Documents ~
$ popd  # pop a directory off the stack
~
$ pwd
/home/ichimonji10
```

These are all shell builtins:

```console
$ type pushd popd dirs
pushd is a shell builtin
popd is a shell builtin
dirs is a shell builtin
```

ci/jjb/jobs/docs.yaml
```diff
@@ -64,3 +64,114 @@
...
+
+            # Add github.com as a known host
+            echo "github.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ==" >> /home/jenkins/.ssh/known_hosts
+            echo "docs-pulp.rhcloud.com,52.1.83.65 ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAwUPkUQ84FKIWhOxy5RBBuR9gbrov2epARmFmaxD7NFRayobnDvl8GaBTbH1kxaZ/tYQeIqVE1assL74KArMQBzz6rj0FEWf0zrXxAY55EGswmWEEdqlYu1LbIxDCG6opqkiq6ocxjea9K3XYq+2aYoAvI3sshSImTYZP1glFhrh3QUsNJHOfDboTLJFNSdLjzXubRMa4eEx7s9pe9kwBOaLNIiVeGzUWg5+KaykSg2UMB3aG127t8kX+OhDYceVR42ehQJ0MjQGlGoNtldxGrlX8NjxUqvJAo6pqNqRK8Cps7/x/m0GPXWAgSZymhurXmj1o2LP5nKLtVzMPvwMb0w==" >> /home/jenkins/.ssh/known_hosts
+
+            chmod 644 /home/jenkins/.ssh/known_hosts
+
+            # create a virtualenv in which to install packages needed to build docs
+            virtualenv -p /usr/bin/python3 --system-site-packages ~/docs_ve
```
Ichimonji10: Python 3 ships with the `venv` module in the standard library. It is preferable to use that instead of the third-party virtualenv package when creating virtualenvs, unless support for Python 2 is needed. Sample usage: `python3 -m venv ~/docs_ve`.

ci/jjb/jobs/docs.yaml
```diff
@@ -64,3 +64,114 @@
...
+              --connection local
+
+            # Install drf_openapi to build REST API docs
+            ansible all \
+                --inventory localhost, \
+                --connection local \
+                --become \
+                --become-user pulp \
+                -m pip \
+                -a 'virtualenv=~/pulpvenv name=git+https://github.com/limdauto/drf_openapi.git@54d24fb#egg=drf_openapi'
```
Ichimonji10: If the commit hash `54d24fb` is intentional, an explanatory comment would be nice.

ci/jjb/jobs/docs.yaml
```diff
@@ -1,5 +1,5 @@
 # This set of Jenkins jobs which build the Pulp platform and plugin documentation and publishes
-# it to https://docs.pulpproject.org/
+# it to https://docs.pulpproject.org/ for pulp2
```
Ichimonji10: :+1:

ci/jjb/jobs/docs.yaml
```diff
@@ -64,3 +64,114 @@
...
+        - timeout:
+            # Timeout in minutes
+            timeout: 240
+            timeout-var: 'BUILD_TIMEOUT'
+            fail: true
+    builders:
+        - shell: |
+            #!/bin/bash
+            # install packages deps required to build docs
+            sudo dnf install graphviz plantuml python-virtualenv postgresql-devel python3-devel gcc ansible -y
```
Ichimonji10: `python-virtualenv` should be unnecessary. See the comment on making a virtualenv, below.

ci/jjb/jobs/docs.yaml
werwty: Sweet! this is exactly what I'm looking for. Thanks ichi :+1:

werwty:
I don't have a good answer to how we can solve the maintenance issue. Maybe we should go in and refactor/rename some of the jobs to better reflect what they do? Or maybe we should consider ansiblizing more of the shared tasks.

Ichimonji10:
I wouldn't worry about the maintenance issue at the time. It's bigger than this one pull request. The JJB definitions are hard to learn and maintain, and that holds true for *all* of the definitions. I bring up the issue of maintenance now because this PR provides a good example. There's now three different jobs that install and run Pulp 3, and much of the necessary logic is duplicated.

PR #468, opened 2018-01-11
Pins sphinx to 1.6.5

dkliban:


ci/jjb/jobs/docs.yaml
```diff
@@ -51,7 +51,8 @@
             # create a virtualenv in which to install packages needed to build docs
             virtualenv -p /usr/bin/python3 --system-site-packages ~/docs_ve
             source ~/docs_ve/bin/activate
-            pip3 install celery 'django<2' django-filter djangorestframework==3.6.4 djangorestframework-jwt drf-nested-routers psycopg2 sphinx git+https://github.com/snide/sphinx_rtd_theme.git@abfa98539a2bfc44198a9ca8c2f16efe84cc4d26 pyyaml virtualenv
+            pip3 install celery 'django<2' django-filter djangorestframework==3.6.4
```
dralley: Why the newline, and will this be problematic?

Ichimonji10: In editors that are set to not wrap code, this line of code is hard to read and edit as-is. I suspect a backslash is necessary, though.

ci/jjb/jobs/docs.yaml
```diff
@@ -126,7 +127,8 @@
             # create a virtualenv in which to install packages needed to build docs
             python3 -m venv --system-site-packages ~/docs_ve
             source ~/docs_ve/bin/activate
-            pip3 install celery sphinxcontrib-swaggerdoc 'django<2' django-filter djangorestframework djangorestframework-jwt drf-nested-routers psycopg2 sphinx git+https://github.com/snide/sphinx_rtd_theme.git@abfa98539a2bfc44198a9ca8c2f16efe84cc4d26 pyyaml virtualenv
+            pip3 install celery sphinxcontrib-swaggerdoc 'django<2' django-filter
```
Ichimonji10: Please add a backslash here, too.

werwty:
The docs builder issues were fixed via updating the docs. So we do not need to pin sphinx to 1.6.5

PR #471, opened 2018-01-16
Add coreapi and aiohttp to pr-doc dependencies

werwty:


werwty:
My guess is one of our dependencies dropped their dependency on coreapi. We should have this here explicitly since we list it as a dependency in pulpcore/setup.py

ci/jjb/jobs/pr-docs.yaml
```diff
@@ -73,7 +73,7 @@
             # create a virtualenv in which to install packages needed to build docs
             virtualenv -p /usr/bin/python3 --system-site-packages ~/docs_ve
             source ~/docs_ve/bin/activate
-            pip3 install celery sphinxcontrib-swaggerdoc 'django<2' django-filter djangorestframework==3.6.4 djangorestframework-jwt drf-nested-routers psycopg2 sphinx git+https://github.com/snide/sphinx_rtd_theme.git@abfa98539a2bfc44198a9ca8c2f16efe84cc4d26 pyyaml virtualenv
+            pip3 install celery sphinxcontrib-swaggerdoc 'django<2' coreapi aiohttp django-filter djangorestframework==3.6.4 djangorestframework-jwt drf-nested-routers psycopg2 sphinx git+https://github.com/snide/sphinx_rtd_theme.git@abfa98539a2bfc44198a9ca8c2f16efe84cc4d26 pyyaml virtualenv
```
bmbouter: can you also add `aiofiles` to these which is coming with the next push https://pypi.python.org/pypi/aiofiles/0.3.2

PR #472, opened 2018-01-18
Require Fedora 26 or newer for pulp-fixtures

Ichimonji10:
The Pulp Fixtures project requires Bash 4.4 or newer, which is present
in Fedora 26 and newer.

The "or" syntax used by this commit can be seen here:
https://docs.openstack.org/infra/jenkins-job-builder/definition.html#job


Ichimonji10:
Required for https://github.com/PulpQE/pulp-fixtures/pull/72

PR #476, opened 2018-01-23
Job updates for 2.16

pcreech:
Job updates for 2.16


pcreech:
That's actually the only reason I preserved the 'pulp_version' variable.  That environment variable is used in the scripts that are referenced and expects a legitimate pulp x.y version.

PR #490, opened 2018-02-16
Install pulp-file as part of pulp-3-dev-{os} jobs

Ichimonji10:
The Pulp 3 testing jobs currently only install pulpcore. No plugins are
installed. Change this state of affairs by also installing the pulp-file
plugin.

With both pulpcore and pulp-file installed, all of the Pulp Smash tests
for Pulp 3 can run.


Ichimonji10:
Tested under a slightly different job name.

PR #492, opened 2018-02-26
Name Pulp 3 testing job after install method

Ichimonji10:
The `pulp-3-dev-{os}` job installs Pulp 3 from PyPI packages, and then
tests it with Pulp Smash. The "dev" in its name seems arbitrary and
meaningless. The job certainly isn't installing a development version of
Pulp 3. Rename the job to `pulp-3-pypi-{os}`, so that the name reflects
which type of installation is at hand.


Ichimonji10:
Tested by slightly modifying this PR and creating jobs named `jaudet-pulp-3-pypi-*`. You can go on to Jenkins and see the results. (Green.)

PR #496, opened 2018-03-02
Removes Pulp PR docs builder job

dkliban:


ammaritiz:
Remove [here](https://github.com/pulp/pulp-ci/blob/master/ci/jjb/jobs/projects.yaml#L238) too?

dkliban:
@ammaritiz good catch ... i updated the PR

PR #497, opened 2018-03-15
Pulp 2.16 Job updates

pcreech:


ci/jjb/jobs/projects.yaml
```diff
@@ -48,14 +37,15 @@
...
       - 'f27'
       - 'rhel7'
     build_version:
       - master:
-          pulp_version: '2.16'
+          pulp_version: '2.17'
           reverse_trigger: 'master'
+      - 2.16:
+          pulp_version: '2.16'
       - 2.15:
```
bmbouter: when do you think is the right time to remove these?

pcreech: @bmbouter I think when we no longer are planning on doing a .z stream. 

We have typically done y.z up untill y+1.1 is released, then no more .y z streams

bmbouter: Makes sense, thanks!

PR #498, opened 2018-03-16
Test Pulp 3 devel on RHEL 7

Ichimonji10:
Expand the `pulp-3-devel-{os}` job matrix, so that RHEL 7 is also
tested.

### DIFF
--- a/ci/jjb/jobs/projects.yaml
+++ b/ci/jjb/jobs/projects.yaml
@@ -67,6 +67,7 @@
     os:
         - f26
         - f27
+        - rhel7
     jobs:
         - pulp-3-devel-{os}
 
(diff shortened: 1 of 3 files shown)


Ichimonji10:
Tested by creating jobs named `jaudet-pulp-3-devel-{os}` and a script named `jaudet-pulp-3-devel-smasher` on Jenkins and executing them. There are a couple test failures, but there's a couple test failures in the existing jobs too. I suspect a feature changed in Pulp 3.

ammaritiz:
The recent [build](https://pulp-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/jaudet-pulp-3-devel-rhel7/10/console) is complaining about enabling some of the official repos. Not sure though, if it is because of this PR.

Ichimonji10:
I'm not sure what's up with that. I just re-triggered the job, and it succeeded. I suspect the failure is due to subscription-manger being unreliable.

preethit:
@Ichimonji10 Has this been merged?

PR #512, opened 2018-04-23
Adding the index page from the actual docs

bmbouter:


bmbouter:
@asmacdo I agree. The separate index is how it used to work. I think we need to start with a much-simplified script and have Travis do it 100%. We can enable nightly building of the docs. The other thing we could do is go back to read the docs.

PR #514, opened 2018-04-24
Update docs for beta build

werwty:


ci/config/releases/3.0-dev.yaml
```diff
@@ -9,4 +9,4 @@ repositories:
     external_deps: deps/external_deps.json
     git_url: git@github.com:pulp/pulp.git
     git_branch: 3.0-dev
-    version: 3.0.0-0.0.alpha
+    version: 3.0.0-0.0.beta
```
werwty: The release number isn't actually used for anything in pulp3. so it's ok to leave at 0.0

PR #518, opened 2018-05-01
Updated jobs to newer versions of Fedora

dralley:
I fully accept that we might want to be more conservative about upgrading the jobs, I just wanted to highlight all of them at first.

Let me know if I should remove any of these changes - happy to do so.

### DIFF
--- a/ci/jjb/jobs/projects.yaml
+++ b/ci/jjb/jobs/projects.yaml
@@ -39,8 +39,8 @@
 - project:
     name: pulp-dev
     os:
-      - 'f26'
       - 'f27'
+      - 'f28'
       - 'rhel7'
     build_version:
       - master:
@@ -57,8 +57,8 @@
 - project:
     name: pulp-3-pypi
     os:
-        - f26
         - f27
+        - f28
     jobs:
         - pulp-3-pypi-{os}
 
@@ -114,23 +114,19 @@
 - project:
     name: pulp-upgrade
     os:
-        - f24
-        - f25
+        - f27
+        - f28
         - rhel6
         - rhel7
     pulp_version:
-        - 2.10-stable
         - 2.11-stable
         - 2.12-stable
         - 2.13-stable
         - 2.14-stable
         - 2.15-stable
     upgrade_pulp_version:
-        - 2.15-beta
         - 2.16-beta
     exclude:
-        - os: f25
-          pulp_version: 2.10-stable
         - os: rhel6
           upgrade_pulp_version: 2.15-beta
     jobs:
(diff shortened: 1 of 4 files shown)


ci/jjb/jobs/projects.yaml
```diff
@@ -114,23 +114,19 @@
 - project:
     name: pulp-upgrade
     os:
-        - f24
-        - f25
+        - f27
+        - f28
         - rhel6
```
dralley: Why RHEL6 ?

ci/jjb/jobs/projects.yaml
```diff
@@ -114,23 +114,19 @@
...
         - rhel7
     pulp_version:
-        - 2.10-stable
         - 2.11-stable
         - 2.12-stable
         - 2.13-stable
         - 2.14-stable
         - 2.15-stable
     upgrade_pulp_version:
-        - 2.15-beta
```
dralley: Should I leave this here?

ci/jjb/jobs/redmine.yaml
```diff
@@ -4,7 +4,7 @@
 - job:
     name: 'redmine-bugzilla-automation'
     defaults: ci-workflow-runtest
-    node: 'f25-np'
```
dralley: I'm not sure if this can be moved to RHEL, but if it can, it probably should

Ichimonji10:
Is there any specific reason to drop Fedora 26 at this time? It should be supported until August of this year.

dralley:
@Ichimonji10 According to the (N + 2) + 1 month schedule, it should go EOL next month?

Ichimonji10:
Ah, sorry. Yes, I was assuming the typical 13 month support time. I'm likely wrong.

This change looks fine to me. I'm happy to merge, push to Jenkins, and see what happens. If we're wrong, we can easily revert. Perhaps we should wait until @pcreech finishes his Crane release, though.

pcreech:
Crane beta is out.  Ga is Thursday.

I'm assuming fedora 28 has been added to nodepool?

We still don't have build infra (Koji) for f28 yet due to higher priorities, but you should be able to use f27 packages for the time being.  I'll push harder for f28 koji next week

Ichimonji10:
See: https://github.com/pulp/pulp-ci/pull/527#issuecomment-391831180

Ichimonji10:
For some reason, nodepool is having trouble generating Fedora 28 images. That needs to be fixed before this can be merged. For more information, log in to the nodepool host and look at `journalctl -u nodepoold --since=...`.

dralley:
@Ichimonji10 what's the status on F28 for Nodepool?

PR #532, opened 2018-05-30
Update obal to at least fail pulp-packaging

pcreech:


pcreech:
@asmacdo Yeah, i'll be evalutaing that, but there also might be a 0.0.4 that will be pushed depending on findings.  For the time being, this gets the immediate result of causing the jenkins job to fail instead of pass

PR #533, opened 2018-05-30
switches master branch to 2-master

dkliban:


ci/jjb/jobs/projects.yaml
```diff
@@ -162,7 +162,7 @@
       - pulp_deb:
           min_coverage: 100
           unittest_branches:
-            - master
+            - 2-master
```
bmbouter: Do we want this one?

PR #557, opened 2018-07-12
Don't use Fedora 26 for testing Pulp 2 or 3 master

Ichimonji10:
Fedora 26 is EOL. Thus, it shouldn't be used for Pulp 2 master or Pulp 3
master development.

See: https://github.com/pulp/pulp-ci/pull/556


Ichimonji10:
I might totally misunderstand what we support for development, so a review from a developer would be great.

PR #559, opened 2018-07-12
Use command module instead of shell module

Ichimonji10:
The shell module performs transformations that can be non-obvious.


rochacbruno:
LGTM

PR #560, opened 2018-07-12
Use package module instead of action module

Ichimonji10:
The "package" module is well-documented and well-understood. Action
plugins are a more advanced feature. Why use an advanced feature when a
more well-understood feature will do?

As a bonus, the resulting code is in standard YAML syntax, instead of
being a single long string that must be interpreted by Ansible.


rochacbruno:
LGTM

PR #563, opened 2018-07-19
Let pulp-smash-runner job give debug info

Ichimonji10:
This will help solve a current test runner failure, and help with
debugging future test runner failures.


Ichimonji10:
I think the current failure is because the `pulp-2-tests` role creates a virtualenv in `/home/jenkins/.virtualenvs/...`, and the subsequent Jenkins jobs references `~/.virtualenvs/...`, which resolves to `/var/lib/jenkins/.virtualenvs/...`. This kind of debugging information will let me make fixes with more certainty.

PR #566, opened 2018-07-20
Don't install Pulp Smash from source

Ichimonji10:
The pulp-2-tests role installs Pulp 2 Tests. Pulp 2 Tests depends on a
specific range of Pulp Smash versions, and thus will pull in the correct
version as needed. As a result, installing Pulp Smash directly from git
an unnecessary extra step. In fact, installing Pulp Smash from source is
harmful: it prevents us from verifying that Pulp 2 Test's dependencies
are correctly specified.

See: https://github.com/PulpQE/Pulp-2-Tests/pull/3

### DIFF
--- a/ci/ansible/roles/pulp-2-tests/tasks/main.yml
+++ b/ci/ansible/roles/pulp-2-tests/tasks/main.yml
@@ -67,12 +67,6 @@
     state: latest
   when: result is changed
 
-- name: Install Pulp Smash
-  pip:
-    virtualenv: '{{ ansible_user_dir }}/.virtualenvs/pulp-2-tests'
-    name: git+https://github.com/PulpQE/pulp-smash.git#egg=pulp-smash
-    state: present
-
 - name: Install Pulp 2 Tests
   pip:
     virtualenv: '{{ ansible_user_dir }}/.virtualenvs/pulp-2-tests'


Ichimonji10:
https://github.com/PulpQE/Pulp-2-Tests/pull/3 should be merged before this.

rochacbruno:
I am not so sure about this, 

If we change this then we are going to install pulp-smash` from PyPI which is released once a week on Thursdays.

If we are in the process of a release, let's say in a Monday and we need to fix or add something we will block the release and require someone to release to PyPI before the builds.

I suggested the pulp 3 tests to install pulp-smash from github master as it is having the fresh code with mostly of the changes and fixes addressed quickly.

Why not do the same for Pulp 2 tests here?

I would keep installing from github source as we have a good code review process and we only merge code there if tested and passing.

rochacbruno:
If we proceed to install from PyPI maybe we may need to setup a nighly build/release for pulp-smash and then use `--pre` release process.

https://packaging.python.org/tutorials/installing-packages/#installing-prereleases 

Which do not differ too much on having it installed from github.

Ichimonji10:
Fair point. I'll close this, but I'd like to expand the test matrix in Pulp 2 Tests. Specifically, I'd like for Pulp 2 Tests to test against both Pulp Smash on PyPI and Pulp Smash from source.

Ichimonji10:
The flip side of this is that installing Pulp Smash from source prevents us from making breaking changes to Pulp Smash itself. One of the benefits of separating an application from the library it depends on is that, well, the application and the underlying library are separate. Loose coupling is a Good Thing.

PR #567, opened 2018-07-20
Adding disk image builder jobs for building opensstack images

ragabala:
This jobs will be creating new images on the fly in the openstack using
the glance Apis. These jobs will be triggered in the Image building
machine and is used for replacing existing images with new ones.

### DIFF
--- a/ci/jjb/jobs/disk-image-builder.yaml
+++ b/ci/jjb/jobs/disk-image-builder.yaml
@@ -1,0 +1,20 @@
+- job-template:
+    name: 'disk-image-builder-base'
+    node: 'disk-image-builder'
+    properties:
+        - qe-ownership
+    scm:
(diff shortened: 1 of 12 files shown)


ci/jjb/jobs/disk-image-builder.yaml
```diff
@@ -0,0 +1,24 @@
+- job-template:
+    name: 'disk-image-builder-base'
+    node: 'disk-image-builder'
+    properties:
+        - qe-ownership
+    scm:
+      - git:
+          url: https://github.com/ragabala/pulp-ci.git
```
dkliban: this should point to pulp/pulp-ci

ragabala: Made the change @dkliban

PR #569, opened 2018-08-01
Make variable expansions safer

Ichimonji10:
Do the following in several newly-added Bash scripts:

* Surround all variable expansions with double quotes.
* Use curly-brace syntax for variables.

What does this look like in practice? Code like `my/$variable/path` is
changed to `"my/${variable}/path"`.

Together, these two changes make variable expansions much safer. Using
quotes ensures that Bash won't do things like split expanded variables
based on whitespace they contain. And using curly-brace variable syntax
avoids ambiguity surrounding which characters might or might not be
included in a variable name.


Ichimonji10:
@ragabala

PR #571, opened 2018-08-03
Remove Pulp 3 Jenkins Jobs

dralley:


Ichimonji10:
This is being done because the Pulp 3 functional tests are now run by upstream. QE will re-introduce similar jobs that run Pulp 3 functional tests on RHEL, but I think it's a good idea to do this from scratch after an installer of some kind is available.

PR #573, opened 2018-08-06
Purge old jjbs

ragabala:
This commit will remove all the old jjb from the pulp ci. But the jenkins jobs has to be removed manually

### DIFF
--- a/ci/jjb/jobs/disk-image-builder.yaml
+++ b/ci/jjb/jobs/disk-image-builder.yaml
@@ -5,6 +5,8 @@
         - qe-ownership
     scm:
         - pulp-ci-github
+    triggers:
+        - timed: 'H H * * 6'
     wrappers:
         - config-file-provider:
             files:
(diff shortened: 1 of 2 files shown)


ci/jjb/jobs/disk-image-builder.yaml
```diff
@@ -5,6 +5,8 @@
         - qe-ownership
     scm:
         - pulp-ci-github
+    triggers:
+        - timed: 'H H * * 6'
```
rochacbruno: Is this being replaced by https://github.com/pulp/pulp-ci/pull/567 ?

ragabala: Hi @rochacbruno ,

#567 was merged without scheduling disk image builder job. This commit will do that. The jobs will be scheduled to run every Saturday replacing the existing images on OpenStack.
 
Do let me know if this answers your question.

PR #581, opened 2018-08-14
Added `attr` to required packages list

rochacbruno:
Attr is required to test SELinux, so adding it to the required packages list.


kersommoura:
@rochacbruno, thank you.

PR #583, opened 2018-08-14
Adjust version_compare to version using filters.

kersommoura:
Adjust version_compare to version.

Requires:ansible is version('2.5', '>=')

See: https://docs.ansible.com/ansible/2.6/user_guide/playbooks_tests.html#version-comparison


kersommoura:
It was used to install Pulp Version: 2.17b1 on a RHEL machine.

PR #584, opened 2018-08-15
Modernize the pulp role's tasks file

Ichimonji10:
Do the following:

* Use the "package" module instead of the more-abstract "action" module.
  This transforms code like this:

  ```yaml
  action: "{{ ansible_pkg_mgr }} name=python-django state=present
  ```

  To code like this:

  ```yaml
  package: name=python-django state=present
  ```

* Transform code from the ansible-specific inline syntax to yaml syntax.
  This changes code like this:

  ```yaml
  package: name=python-django state=present
  ```

  To code like this:

  ```yaml
  package:
    name: python-django
    state: present
  ```

* Drop an outdated hack intended to work around a weird interaction
  between Fedora 26 and Ansible 2.4.0. See:
  https://github.com/ansible/ansible/issues/26868
* Use the "command" module instead of the "shell" module where possible.
  Bash syntax is quirky and often badly understood, and this change
  prevents an entire class of bugs.
* Don't call "sudo" in "command" tasks. All tasks are already being run
  as root. And if one wishes to run a task as root, the better way to do
  it is by adding `become: true` to a task.
* Make sure that dependencies are "present," not "latest." This is
  necessary for idempotence (very much the Ansible way!) and greatly
  improves runtime upon subsequent playbook runs.
* When installing multiple packages, pass a list of packages to be
  installed, instead of looping through each package to be installed
  one-by-one. This allows the package manager to optimize its operations
  by doing things like resolving dependncies just once, running
  post-transaction hooks just once, and so on.

### DIFF
--- a/ci/ansible/roles/pulp/tasks/main.yaml
+++ b/ci/ansible/roles/pulp/tasks/main.yaml
@@ -1,5 +1,6 @@
 ---
 - block:
+
   - name: Open firewall ports (iptables)
     iptables:
       action: insert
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Tested against Fedora 27 and RHEL 7 with Pulp 2.17.0 beta.

PR made because I was looking into the issue wherein Pulp 2.17 can't be installed on F28, and my fingers just kept on moving. :wink:

Ichimonji10:
@kersommoura You were working with Ansible recently. You might enjoy looking at this.

ci/ansible/roles/pulp/tasks/main.yaml
```diff
@@ -9,41 +10,54 @@
       destination_port: "{{ item.port }}"
       jump: ACCEPT
     with_items:
```
rochacbruno: looks like ` with_items`  is deprecated in favor of ` loop`  https://docs.ansible.com/ansible/2.5/user_guide/playbooks_loops.html

PR #586, opened 2018-08-16
Adding debuggers for master-dev jobs in CI

ragabala:
The nightlies are failing to run because of dns resolution. Will help if
the actual hostnames of the slaves are known during such failures. This
PR will help in logging the `hostname` and `hostname --all-fqdn` on the
slave system.

### DIFF
--- a/ci/jjb/jobs/macros.yaml
+++ b/ci/jjb/jobs/macros.yaml
@@ -52,6 +52,9 @@
     # This hostname can be connected from other open stack vms.
     builders:
         - shell: |
+            # Hostnames that are set
+            echo "$(hostname --all-fqdn)"
+            echo "$(hostname)"
             # Setting the hostname of the system to enable connectivity
             hostname="$(hostname --all-fqdn | grep openstacklocal | head -n 1)"
             hostname="$(echo ${hostname})"


ci/jjb/jobs/macros.yaml
```diff
@@ -52,6 +52,9 @@
     # This hostname can be connected from other open stack vms.
     builders:
         - shell: |
+            # Hostnames that are set
+            echo "$(hostname --all-fqdn)"
+            echo "$(hostname)"
```
Ichimonji10: The `hostname` executable already prints to stdout. There's no benefit to capturing its output and then printing to stdout with `echo`.

ragabala: Hi @Ichimonji10 ,

The hostname which gets printed are ones which are filtered out by the `grep openstacklocal` command. The actual output which shows the `hostname --all-fqdn` is not seen in the Jenkins jobs. Let me know if I am off base here.

Thanks

Ichimonji10: I think my point has been lost in translation. See: https://github.com/pulp/pulp-ci/pull/587

PR #588, opened 2018-08-17
Adding F28 python3 firewalld bindings

ragabala:
This PR addresses two issues that are related to Fedora28.
* It makes sure Python3 interpreter is used for running Ansible in Fedora 28. This was required since the `python2-firewall` packages are not used in F28.
* SELinux is set in the **permissive** mode in F28, Since the tasks that starts squid.service [1] fails in F28.

[1] https://pulp.plan.io/issues/3926

### DIFF
--- a/ci/jjb/jobs/projects.yaml
+++ b/ci/jjb/jobs/projects.yaml
@@ -41,8 +41,9 @@
     - 2.16:
         pulp_version: '2.16'
     os:
-    - 'f27'
-    - 'rhel7'
+    - f27
+    - rhel7
+    script-var: ''
     reverse_trigger: '{build_version}-dev'
     jobs:
     - pulp-{build_version}-dev-{os}
@@ -57,13 +58,16 @@
         pulp_version: '2.18'
         reverse_trigger: '2-master'
     os:
-    - 'f27'
-    - 'f28'
-    - 'rhel7'
-    - 'rhel7-fips'
+    - f27
+    - f28:
+        script-var: '-f28'
+    - rhel7
+    - rhel7-fips
     reverse_trigger: '{build_version}-dev'
+    script-var: ''
     jobs:
     - pulp-{build_version}-dev-{os}
+
 
 - project:
     name: disk-image-builder-base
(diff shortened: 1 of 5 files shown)


ci/jjb/jobs/projects.yaml
```diff
@@ -41,8 +41,9 @@
     - 2.16:
         pulp_version: '2.16'
     os:
-    - 'f27'
-    - 'rhel7'
+    - f27
+    - rhel7
+    script-var: ''
```
Ichimonji10: What does `script-var: ''` do?

ragabala:
@Ichimonji10 @kersommoura @rochacbruno 

I think the travis is failing because of older version of jenkins-job util. I am running 2.0.2 and the JJB tests runs fine in the local. I was also able to update the jobs from my local to jenkins. Let me know if there is a way to update jenkins-jobs in travis

Ichimonji10:
To update Travis, edit `.travis.yml`.

ci/jjb/scripts/pulp-install-f28.sh
```diff
@@ -0,0 +1,12 @@
+setenforce permissive
```
kersommoura: @ragabala, perhaps just skip just the failing test due to selinux issues can be another solution.

ci/jjb/jobs/projects.yaml
ragabala: script-var is a place-holder for changing the files according to the OS .

If the OS is anything else other than **f28**, the pulp-dev projects load the script pulp-install.sh.
If it is equal to f28, the pulp-dev projects load pulp-install-f28.sh ( which has SELinux permissive and also ansible py interpreter=3).
Thus the scripts are loaded dynamically

ci/jjb/scripts/pulp-install-f28.sh
ragabala: @kersommoura: The issue is not in tests .. but in the installation of pulp. The pulp-ci fails to install pulp as squid fails to start in the lazy role if SELinux is enabled in F28.

PR #589, opened 2018-08-17
Install Python 2 on Fedora

Ichimonji10:
Pulp 2 requires Python 2.


Ichimonji10:
Someone with a Fedora host should be able to run the installer against it like so:

```bash
ansible-playbook pulp_server.yaml \
  --inventory fedora-27,fedora-28 \
  -e 'ansible_python_interpreter=/usr/bin/python3'
```

I've let off a stream of curses on many, many occasions about this deficiency.

PR #606, opened 2018-09-28
Migrating ``disk-image-builder-base`` job to Open Stack node.

ragabala:
The current ``disk-image-builder-base`` job is running on a physical
instance. Migrating this job to be run on open stack gives much more
availability. As a part of this commit, a new instance called
``disk-image-builder-os`` is set up in the jenkins-openstack cloud
plugin which points to a fedora 27 instance in openstack. Once this job
runs, the fedora-27 instance gets provisioned which will act as a host
for disk image building.

### DIFF
--- a/ci/jjb/jobs/disk-image-builder.yaml
+++ b/ci/jjb/jobs/disk-image-builder.yaml
@@ -1,6 +1,6 @@
 - job-template:
     name: 'disk-image-builder-base'
-    node: 'disk-image-builder'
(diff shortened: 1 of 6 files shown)


kersommoura:
@ragabala, good job! Let`s document those changes. Explain any extra detail that we will help another person to update the base image used to create other images. For instance, in case we need to update distro version and so on.

kersommoura:
@rochacbruno, any comments?

PR #613, opened 2018-10-31
Adding python wrapper for getting latest openstack image.

ragabala:
This commit adds a new python executable which will be called by the
``builder.sh`` file. The executable is responsible for getting an input
image name, finding all the images in openstack matching the name, and
getting the latest image in terms of ``updated_at``.


kersommoura:
Thanks @ragabala. Goob job.

PR #616, opened 2018-11-08
add newer versions to pulp_installer job

rochacbruno:


ci/jjb/jobs/pulp-installer.yaml
```diff
@@ -35,6 +35,8 @@
         - choice:
             name: PULP_VERSION
             choices:
+                - '2.19'
+                - '2.18'
                 - '2.17'
                 - '2.16'
```
kersommoura: 2.16 is not supported anymore. I think is fine to remove from here.

PR #623, opened 2018-12-11
Update pulp-fixtures trigger to build on Saturday only

bherrin3:
Part of the simplification of ci jobs.

The group decided the job only needs to be ran weekly. Matching
the usage pattern to be similar to disk-image-builder. Jenkins will
load balance the job start time on Saturday with the `H H'.

There is no ticket associated with this change other than a group discussion.

### DIFF
--- a/ci/jjb/jobs/pulp-fixtures-publisher.yaml
+++ b/ci/jjb/jobs/pulp-fixtures-publisher.yaml
@@ -13,7 +13,7 @@
                 - '*/master'
             skip-tag: true
     triggers:
-        - timed: '@midnight'
+        - timed: 'H H * * 6'
     wrappers:
         - jenkins-ssh-credentials
     builders:


ci/jjb/jobs/pulp-fixtures-publisher.yaml
```diff
@@ -13,7 +13,7 @@
                 - '*/master'
             skip-tag: true
     triggers:
-        - timed: '@midnight'
+        - timed: 'H H * * 6'
```
rochacbruno: Is `H H` the same as `0 0` or is it the `pick a random time` stuff?

bherrin3: @rochacbruno from the documentation (I think) on github:

> To allow periodically scheduled tasks to produce even load on the system,
    the symbol <code>H</code> (for “hash”) should be used wherever possible.
    For example, using <code>0 0 * * *</code> for a dozen daily jobs
    will cause a large spike at midnight.
    In contrast, using <code>H H * * *</code> would still execute each job once a day,

I am not tied to the current implementation. Going with something like;

```suggestion
        - timed: '0 0 * * 6'
```

Is fine with me. Since I don't own Jenkins, I was attempting to use the load distribution I believe Jenkins provides.

Source: https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/triggers/TimerTrigger/help-spec.jelly

bherrin3: Just for the conversation:

> In addition, <code>@yearly</code>, <code>@annually</code>, <code>@monthly</code>,
    <code>@weekly</code>, <code>@daily</code>, <code>@midnight</code>,
    and <code>@hourly</code> are supported as convenient aliases.
    These use the hash system for automatic balancing.
    For example, <code>@hourly</code> is the same as <code>H * * * *</code> and could mean at any time during the hour.
    <code>@midnight</code> actually means some time between 12:00 AM and 2:59 AM.

We were already using balancing with `@midnight` which runs within a range. Therefore, if we haven't had problems thus far, I don't see a real issue with `H H * * 6`.

PR #628, opened 2018-12-14
Rename CCRC to GPLCC in COMMITMENT

CodeHeeler:
The "Common Cure Rights Commitment" is being renamed to "GPL Cooperation Commitment".
As part of our effort to promote this initiative further we want to standardize on that name (GPLCC).

ref #4153
https://pulp.plan.io/issues/4153


ggainey (approved):
Not sure how this didn't get merged forever ago - but the second-best time to do a thing is "now". Approved.

PR #629, opened 2018-12-19
Remove workaround python-django issue untill fixed

kersommoura:
This workaround was introduced due to an error to install
`python-django16-1.6.11.6-7.el7`. This package is now in a newer
version. Remove this workaround.

Current python-django version available using epel: 1.11.13-4.el7o

See: https://pulp.plan.io/issues/4281


pcreech:
This should be removed, ACK

PR #630, opened 2019-01-02
Adding Safe delete logic to disk image builder

ragabala:
The current implementation of Disk Image Builder deletes an image in the
upstream without considering the status of the newly created image. Thus
even before the new image created is uploaded, the old image is deleted
. This was done to remove any name conflicts, however this creates a
threat of removing a working image in the case of failure. This commit
takes care of preserving the old image when the latest build fails.

### DIFF
--- a/ci/open_stack_plugin/disk_image_builder/scripts/builder.sh
+++ b/ci/open_stack_plugin/disk_image_builder/scripts/builder.sh
@@ -89,17 +89,10 @@ download_base_image(){
 
 remove_existing_image(){
 # remove existing images in OpenStack
-# param $1: OS name
-# param $2: OS version
-# param $3: OS identifier : `fips` or `common`
+# param $1 : Image Id to be deleted
 
-    echo "removing ${1}_${2}_${3}_DIB_updated"
-    value="${1}_${2}_${3}_DIB_updated"
-    get_image_id_from_name "${value}"
-    if [[ "${_image_id_temp}" ]];then
-        echo "deleting existing image ${_image_id_temp}"
-        glance image-delete "${_image_id_temp}"
-    fi
+    echo "deleting existing image ${1}"
+    glance image-delete "${1}"
 }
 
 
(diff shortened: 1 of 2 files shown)


bherrin3:
## tl;dr - This is good. Thanks!

@ragabala logically, I agree with moving this logic out of the remove and checking. Thanks!

The global value for `_image_id_temp` is getting mutated in `get_image_id_from_name()`. From purely a future troubleshooting perspective, it might be better to handle this as a `local` and echo the value (proper testing of course)

This is by no means a `do it` statement. What you have works and I don't want to over-complicate what we are doing. **_Maybe we can look into this in another PR or I could submit a PR for this, since I believe this is out of scope for this fix._**

## Just some other information 

We could scope locally:
```
get_image_id_from_name(){
# Retrieve image ID for a given image name from OpenStack.
# Since many images can have the same name, We select the id from one of the images.
# We use the ``./py_get_image_id`` python executable for getting the image id.
# param $1: Image_name
    local _image_id_temp="$(${scripts_dir}/py_get_image_id ${1})"
    echo $_image_id_temp

}
``` 
Rather than a mutated global. 
This is a nitpick mostly because bashr's`return` is a `0-255` return code and not a value return. Bash functions are really designed to return output streams.

Therefore, changing the original call from:
```
    get_image_id_from_name "${value}"
    prev_image="${_image_id_temp}"
```

To something that creates a sub shell, `${..}`, that will not affect the parent shell variables might be preferable for future investigation. 

I would need to test the exact syntax (as this is bash), but I would imagine something like:
```    
    prev_image=${get_image_id_from_name "${value}"}
```

That way, the local function value for `_image_id_temp` is evaluated in a sub-shell but the stdout is still returned to `$prev_image` without overloading a global variable (and bad-stuff happening).

## Example
Here is a more overt example of using the `local` in a function and then calling a function execution in a sub-shell that won't affect the value of a same-name (bad naming standards) global variable `__shoki`.

```
#!/bin/bash

# This is the name of my dog, Shoki
# By default, he wants to bark 45 times at everything.
__shoki=45

# This is how many cars he sees
cars=5

# 
dogs_brain()
{
	local __shoki=`echo -e "Shoki bark at $1 cars!!\n"`
	echo $__shoki
}

family_hears="$(dogs_brain "$cars")"
echo -e "The family heard $family_hears\n"
echo -e "The global value of __shoki wanitng to bark is still $__shoki\n"
```
Returning:

```
[herring@redherring test]$ ./functionTest.sh 
The family heard Shoki bark at 5 cars!!

The global value of __shoki wanting to bark is still 45

[herring@redherring test]$
```

##  References
* [Good Stack Overflow 1](https://stackoverflow.com/questions/23564995/how-to-modify-a-global-variable-within-a-function-in-bash)
* [Another Good Stackoverflow answer on return vs local 2](https://stackoverflow.com/questions/17336915/return-value-in-a-bash-function)
* [Good Stack Overflow 3](https://unix.stackexchange.com/questions/217600/returning-a-variable-from-a-function)
* [Bash Function Primer](http://blog.joncairns.com/2013/08/what-you-need-to-know-about-bash-functions/)
* [TPLD Advanced Bash Scripting Guide Link](https://www.tldp.org/LDP/abs/html/complexfunct.html)
* [Linux Journal Article on Returning Values](https://www.linuxjournal.com/content/return-values-bash-functions)

bherrin3:
:+1:

ragabala:
Thanks @bherrin3 ,

Those are wonderful suggestions. Thanks for nicely putting them.
Will work on another PR correcting this.

PR #632, opened 2019-01-25
add fedora 29 image

rochacbruno:
Adding Fedora 29 image to have an updated OS for Pulp 3 Ci after having some problems running Junit reports on Fedora 28.


kersommoura:
Merged.

PR #633, opened 2019-01-25
Add pulp3 ci

rochacbruno:
Pulp 3 - Master added to jenkins CI

New jobs:
- pulp3-master-dev-f28
- pulp3-master-dev-f29
- pulp3-smash-runner

*NOTE:* The single node pulp-3-devel job left here because it can be useful as an example in the future.

### DIFF
--- a/ci/jjb/jobs/pulp3-dev.yaml
+++ b/ci/jjb/jobs/pulp3-dev.yaml
@@ -1,0 +1,70 @@
+- job-template:
+    name: 'pulp3-{build_version}-dev-{os}'
+    node: '{os}-os'
+    properties:
+        - qe-ownership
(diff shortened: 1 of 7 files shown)


ci/jjb/jobs/pulp3-dev.yaml
```diff
@@ -0,0 +1,70 @@
...
+    properties:
+        - qe-ownership
+        - discard-old-builds
+    scm:
+        - pulp-ci-github
+    wrappers:
+        - config-file-provider:
+            files:
+                - file-id: rhn_credentials
+                  variable: RHN_CREDENTIALS
```
kersommoura: Why do we need those variables?

PR #635, opened 2019-01-31
make pulp-packaging shell step fail if any command failed

evgeni:


pcreech:
ACK

This is more robust than the initial stopgap it supercedes.

PR #641, opened 2019-02-06
Modify priority for pulp repos

kersommoura:
This is a work around to avoid packages provided by package team to
be overwritten by newer versions provided by epel.

See: https://pulp.plan.io/issues/4387

### DIFF
--- a/ci/ansible/roles/pulp/tasks/main.yaml
+++ b/ci/ansible/roles/pulp/tasks/main.yaml
@@ -164,16 +164,6 @@
       priority: 50
   when: pulp_build == "beta" or pulp_build == "stable"
 
-- name: Check fips state
-  shell: cat /proc/sys/crypto/fips_enabled
-  register: fips_state
-
-- name: Install yum-plugin-priorities
-  package:
-    name: yum-plugin-priorities
-    state: present
-  when: fips_state.stdout == '1'
-
 - name: Install Pulp Server
   package:
     name: '@pulp-server-qpid'
--- a/ci/ansible/pulp_server.yaml
+++ b/ci/ansible/pulp_server.yaml
@@ -4,6 +4,7 @@
   roles:
     - role: subscription-manager
       when: ansible_distribution == 'RedHat'
+    - role: pulp-priority
     - role: pulp-fips
       when:
         - enable_fips is defined


ci/ansible/roles/pulp/tasks/main.yaml
```diff
@@ -164,16 +164,6 @@
       priority: 50
   when: pulp_build == "beta" or pulp_build == "stable"
 
-- name: Check fips state
-  shell: cat /proc/sys/crypto/fips_enabled
-  register: fips_state
-
-- name: Install yum-plugin-priorities
```
pcreech: I'm pretty sure this is still needed, as you need yum-plugin-priorities to make priorities work

ci/ansible/roles/pulp/tasks/main.yaml
```diff
@@ -164,16 +164,6 @@
...
-  package:
-    name: yum-plugin-priorities
-    state: present
-  when: fips_state.stdout == '1'
```
pcreech: Instead, you can probably drop this "when" conditional to ensure it's available across the board.

kersommoura:
Tested in a VM box.
```bash
pulp_2_tests/tests/rpm/api_v2/test_service_resiliency.py::MissingWorkersTestCase.test_all ✓                                                                                                         50% █████     
 pulp_2_tests/tests/rpm/api_v2/test_service_resiliency.py::TaskDispatchTestCase.test_all ✓                                                                                                          100% ████

```

ci/ansible/roles/pulp/tasks/main.yaml
```diff
@@ -164,16 +164,6 @@
       priority: 50
   when: pulp_build == "beta" or pulp_build == "stable"
 
-- name: Check fips state
```
pcreech: This might still be useful in the future, so might be worth leaving in.

ci/ansible/roles/pulp/tasks/main.yaml
kersommoura: There was another playbook doing the same. I can modify this, and delete the playbook.

ci/ansible/roles/pulp/tasks/main.yaml
kersommoura: I understood that no matter what box we are installing pulp packages should have priority. I can adjust it.

ci/ansible/roles/pulp/tasks/main.yaml
pcreech: ah, then never-mind, feel free to do the other.

ci/ansible/roles/pulp/tasks/main.yaml
kersommoura: role: pulp-priority does the same. It was a redundancy.

pcreech: Ah, cool.  Then nevermind me.

ci/ansible/roles/pulp/tasks/main.yaml
pcreech: It was more about the check fips state task itself, since it registers a variable to do it.  But not a big deal either way, since currently the priority stuff is the only one that uses it.

pcreech:
ACK

PR #642, opened 2019-02-11
Changing pulp-packaging to run on rhel7-np

ragabala:
The current pulp-packaging job runs on rhel7-os instance which causes it
to fail due to package dependencies. This issue is rectified my making
it to run on rhel7-np instance(Which is an old Rhel 7.5 nodepool
instance). This commit modifies the pulp-packaging
job to run on rhel7-np . This is a temporary fix and will be moved to
rhel7-os once the issue with rhel7-os is fixed.

### DIFF
--- a/ci/jjb/jobs/pulp-packaging.yaml
+++ b/ci/jjb/jobs/pulp-packaging.yaml
@@ -5,7 +5,7 @@
 - job-template:
     name: 'pulp-packaging-{release_config}'
     defaults: ci-workflow-runtest
-    node: 'rhel7-os'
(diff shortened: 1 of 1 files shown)


bherrin3:
@pcreech agreed approved as a temporary fix. Did you expect the base image for rhel7-xx to have subscription manager enabled or do you just need the behavior to be consistent?

With the move to the new OpenStack looming, this base image (rhel7-np) will not be there as it is a custom built image. I would expect the rhel7-os and rhel8-os to be there. This would break on the migration.

PR #644, opened 2019-02-25
Changing Ansible Installer priorities with current 2.18.1 definition

bherrin3:
To solve an EPEL versioning problem with the python-celery stack, the final solution during build has changed the EPOCH on NEVRA for those RPMs to 10:

The resulting logic needed to be updated to:
* use priority on FIPS with EPEL
* do not use priority on non-FIPS with EPEL

The end state of each of those upgrade permutations leave the box in the same RPM state without the end-user to have to play with yum or repository priorities.

Removing the pulp-priority role as it is now deprecated with these changes.

The included, tested changes are the output of the solution for CI.

refs #4387

### DIFF
--- a/ci/ansible/roles/pulp-priority/tasks/main.yml
+++ b/ci/ansible/roles/pulp-priority/tasks/main.yml
@@ -1,5 +1,0 @@
----
-- name: Install yum-plugin-priorities
-  package:
-    name: yum-plugin-priorities
-    state: present
--- a/ci/ansible/pulp_server.yaml
+++ b/ci/ansible/pulp_server.yaml
@@ -1,10 +1,16 @@
 ---
 
 - hosts: all
+  pre_tasks:
+    - name: Check fips state
+      shell: cat /proc/sys/crypto/fips_enabled
+      register: fips_state
+  vars:
+    repo_priority: "{{ '50' if fips_state.stdout == '1'  else '99' }}"
+
   roles:
     - role: subscription-manager
       when: ansible_distribution == 'RedHat'
-    - role: pulp-priority
     - role: pulp-fips
       when:
         - enable_fips is defined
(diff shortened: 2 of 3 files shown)


bherrin3:
@kersommoura thanks for the feedback. I was looking for a better way to convert that role. Good ideas and thanks for giving me some modules and new concepts I haven't worked with yet.

Will re-try with those to minimize and tokenize the priority variable (50 or 99 [default]) to DRY the code up.

ci/ansible/roles/pulp-priority/tasks/main.yml
```diff
@@ -3,3 +3,39 @@
   package:
     name: yum-plugin-priorities
     state: present
+
+- name: Setup Pulp nightly repository
```
kersommoura: @bherrin3, as we talked let`s look for a solution that do not require duplicate the creation of those repos.

bherrin3: @kersommoura updated to be more dry once you let me know about `pre_tasks` ! Neat!

See if you prefer this implementation.

I need to check to see if `pulp-priority` role is being used by anything else. If not, I can remove that file in this PR as well.

bherrin3:
@kersommoura should be ready for real review now. Tested it on fips and non-fips. Did what was expected.

Any other thoughts for optimization I am all for.

Need to check on the `pulp-prioritize` role. Will update/remove if unused.

kersommoura:
@bherrin3, I do not see the role `pulp-prioritize` being used. We cam improve later on if necessary.

bherrin3:
@kersommoura neither did I and I checked the history. I had added that for my initial commit on the priority work. I went ahead and deleted the role as all the logic is now contained within this commit. Nothing has been lost.

PR #646, opened 2019-02-27
Update pulp role to define defaults for repo_priority and evaluate run-time

bherrin3:
Update pulp role to define defaults for repo_priorty and evaluate run-time

As an addendum to PR #644 , the repo_priorty variable in the pulp role must be defined within the scope of the role (not the pulp_server playbook) as other playbooks use that role (broke pulp_backup playbook). 

Making changes to:
- Declare a default of '99' in the pulp/defaults, which is the same as no priority
- Evaluate the proper priority within the role/pulp role for the pulp.repos
- Remove the pre_task declaration from the pulp_server playbook

Retested on fips (99, no plugin) and non-fips (50, plugin) with the pulp_server playbook as well as other playbooks dependent on the pulp role.

refs #4359

### DIFF
--- a/ci/ansible/roles/pulp/tasks/main.yaml
+++ b/ci/ansible/roles/pulp/tasks/main.yaml
@@ -1,4 +1,13 @@
 ---
+# Must check fips state here as pulp-fips could have
+- name: Check fips state
+  shell: cat /proc/sys/crypto/fips_enabled
+  register: fips_state
+
+- name: Set the repo_priority based on fips_state 
+  set_fact:
+    repo_priority: "{{  fips_priority if fips_state.stdout == '1'  else default_priority }}"
+
 # Pulp 2 requires Python 2.
 - name: Install Python 2 (Fedora)
   package:
--- a/ci/ansible/pulp_server.yaml
+++ b/ci/ansible/pulp_server.yaml
@@ -1,13 +1,6 @@
 ---
 
 - hosts: all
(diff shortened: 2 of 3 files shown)


ci/ansible/roles/pulp/tasks/main.yaml
```diff
@@ -137,7 +142,7 @@
       name: Pulp Project repository
       baseurl: "https://repos.fedorapeople.org/pulp/pulp/testing/automation/{{ pulp_version }}/stage/{% if ansible_distribution == 'Fedora' %}fedora-{% endif %}$releasever/$basearch/"
       gpgcheck: 0
-      priority: "{{ repo_priority }}"
+      priority: "{{ '50' if fips_state.stdout == '1' else repo_priority }}"
```
kersommoura: @bherrin3, could we evaluate this just once - probably together with the fips verification, and pass the variable over here?

bherrin3: @kersommoura yup. I didn't know a better way to do it at the time and I didn't like a static value sitting in there for fips.

Updating with the following logic:
- adding defaults for fips_priority, default_priority
- Using `set_fact` after fips eval to dynamically assign the above default to `repo_priority` variable
- Updating the priority statement to be `"{{ repo_priority }}"` above.

Thanks, @rochacbruno ! Re-testing now.

PR #648, opened 2019-03-03
Adding podman configuration to pulp3 smash runner

rochacbruno:


ci/jjb/jobs/pulp3-smash-runner.yaml
```diff
@@ -97,6 +103,22 @@
...
             git clone https://github.com/pulp/pulp_docker.git --branch master
 
+            # Configure podman for docker tests 
+            sudo yum install -y podman
+            sudo bash -c 'cat << EOF > /etc/containers/registries.conf
+            [registries.search]
+            registries = ["docker.io", "registry.fedoraproject.org", "quay.io", "registry.access.redhat.com", "registry.centos.org"]
+
+            [registries.insecure]
+            registries = ["${PULP_SMASH_SYSTEM_HOSTNAME}:8080"]
```
rochacbruno: This line is not evaluating, see `/job/pulp3-smash-runner/83/consoleFull`

I suspect this is not evaluating because the use of `bash -c` creates a subshell where envvars are not defined.

Need to change it or try using `| tee`

rochacbruno:
/job/pulp3-smash-runner/85/consoleFull

ci/jjb/jobs/pulp3-smash-runner.yaml
```diff
@@ -97,6 +103,23 @@
             git clone https://github.com/pulp/pulp_rpm.git --branch master
             git clone https://github.com/pulp/pulp_docker.git --branch master
 
+            # Configure podman for docker tests 
+            sudo yum install -y podman
```
kersommoura: It is a fedora box?
```suggestion
            sudo dnf install -y podman
```

rochacbruno: @kersommoura I would like to keep `yum` as it is agnostic. If someday we use RHEL it will still work. And on fedora `yum` is a kind of `alias` to `dnf`

This solution is already applied to Jenkins and workds see build 85

PR #650, opened 2019-03-19
Update projects.yaml for pulp_deb branches switch

mdellweg:


ci/jjb/jobs/projects.yaml
```diff
@@ -128,7 +128,7 @@
       - pulp_deb:
           min_coverage: 100
           unittest_branches:
-            - master
+            - 2-master
             - 1.6-release
```
mdellweg: I am wondering, whether this should be updated, too...

PR #659, opened 2019-04-11
Change Pulp 3 Default Ports

mikedep333:
https://pulp.plan.io/issues/4556

Content: 8080 -> 24816
API: 8000 -> 24817

[noissue]
Required PR: https://github.com/pulp/pulpcore/pull/75

@rochacbruno Offered to make this update, but I figured I would try anyway. I am a bit surprised; it looks like this repo sometimes was using port 80 for the API, as seen in ci/jjb/jobs/pulp3-smash-runner.yaml

### DIFF
--- a/ci/docs-builder.py
+++ b/ci/docs-builder.py
@@ -141,7 +141,7 @@ def main():
 
     # Get the latest api.yaml file to build the rest api docs
     if is_pulp3:
-        with urllib.request.urlopen("http://localhost:8000/pulp/api/v3/docs/api.yaml") as response, \
+        with urllib.request.urlopen("http://localhost:24817/pulp/api/v3/docs/api.yaml") as response, \
                 open(os.path.join(docs_directory, "api.yaml"), 'wb') as api_file:
             data = response.read()
             api_file.write(data)
--- a/ci/docs/testrubyserver.rb
+++ b/ci/docs/testrubyserver.rb
@@ -3,7 +3,7 @@ require 'webrick'
 include WEBrick
 
 config = {}
-config.update(:Port => 8080)
+config.update(:Port => 24816)
 config.update(:BindAddress => ARGV[0])
 config.update(:DocumentRoot => ARGV[1])
 server = HTTPServer.new(config)
--- a/ci/jjb/scripts/openstack-provision-machine.sh
+++ b/ci/jjb/scripts/openstack-provision-machine.sh
@@ -26,7 +26,7 @@ fi
 
 if ! openstack security group rule list "${KEY_NAME}" | grep -q "0.0.0.0/0"; then
     openstack security group rule create --protocol icmp "${KEY_NAME}"
-    for tcp_port in 22 80 443 5000 5646 5647 5671 8000 8140 8443 9090; do
+    for tcp_port in 22 80 443 5000 5646 5647 5671 24817 8140 8443 9090; do
         openstack security group rule create --protocol tcp --dst-port "${tcp_port}" "${KEY_NAME}"
     done
     for udp_port in 53 69; do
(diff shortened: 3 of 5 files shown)


ci/jjb/scripts/openstack-provision-machine.sh
```diff
@@ -26,7 +26,7 @@ fi
 
 if ! openstack security group rule list "${KEY_NAME}" | grep -q "0.0.0.0/0"; then
     openstack security group rule create --protocol icmp "${KEY_NAME}"
-    for tcp_port in 22 80 443 5000 5646 5647 5671 8000 8140 8443 9090; do
+    for tcp_port in 22 80 443 5000 5646 5647 5671 24817 8140 8443 9090; do
```
rochacbruno: I guess you shoud keep 8000 as the same sec group may be in use by pulp2?

ci/docs/testrubyserver.rb
```diff
@@ -3,12 +3,12 @@
 include WEBrick
 
 config = {}
-config.update(:Port => 8080)
+config.update(:Port => 24816)
```
rochacbruno: is this used only for pulp3 or also  2?

ci/jjb/scripts/openstack-provision-machine.sh
mikedep333: I forgot that we are using pulp-ci for Pulp 2 also.

ci/docs/testrubyserver.rb
mikedep333: Only for Pulp 3.

ci/jjb/scripts/openstack-provision-machine.sh
mikedep333: @rochacbruno Does pulp2 use 8000 at all? I don't see it here: https://docs.pulpproject.org/en/2.18/user-guide/installation/index.html#prerequisites

ci/docs/testrubyserver.rb
mikedep333: I think 8080 was only ever used by Pulp3:
https://docs.pulpproject.org/en/2.18/user-guide/installation/index.html#prerequisites

rochacbruno:
> @rochacbruno Offered to make this update, but I figured I would try anyway. I am a bit surprised; it looks like this repo sometimes was using port 80 for the API, as seen in ci/jjb/jobs/pulp3-smash-runner.yaml

@mikedep333 the use of port 80 is because `nginx` or `httpd` is exposing the internal service using this port, the internal service runs on 8000 but nginx/httpd proxies to 80

ci/docs-builder.py
```diff
@@ -141,7 +141,7 @@ def main():
 
     # Get the latest api.yaml file to build the rest api docs
     if is_pulp3:
```
rochacbruno: @bmbouter is this file being used to build pulp3 docs? (if not can we remove it?)

ci/docs/testrubyserver.rb
```diff
@@ -3,12 +3,12 @@
 include WEBrick
```
rochacbruno: @bmbouter same question here, looks like this file last change was in 2016

rochacbruno:
@mikedep333 I proposed #660 which touches the `pulp3 - master` job related files `pulp-3-smash-runner` and `pulp3-install` so can you dis-include those files from your PR?

I dont know about the other files you changes, so I asked @bmbouter lets see if that files are in use.

Thanks

rochacbruno:
I think this PR can be closed and if further changes are needed a new PR should be created

ggainey:
Closed per @rochacbruno 's comment

PR #661, opened 2019-04-12
Adding rhel7 to Pulp-3-Master matrix, removing f28

rochacbruno:
Part 1:

- Adding `rhel7` to `os` list.

Expected:

- We expect it to fail, lets see exactly how and lets have a job to debug.

Part 2: (coming next)

- This failure will push us to work on it.

### DIFF
--- a/ci/jjb/jobs/projects.yaml
+++ b/ci/jjb/jobs/projects.yaml
@@ -35,8 +35,8 @@
         pulp_version: '3.0'
         reverse_trigger: '3-master'
     os:
-    - f28
     - f29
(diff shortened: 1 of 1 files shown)


ci/jjb/jobs/projects.yaml
```diff
@@ -35,8 +35,8 @@
         pulp_version: '3.0'
         reverse_trigger: '3-master'
     os:
-    - f28
     - f29
+    - rhel7
```
kersommoura: What is the goal to test the installer?

PR #666, opened 2019-04-23
Problem: f27 no longer supported

dkliban:
Solution: remove f27 from unittest matrix

[noissue]

### DIFF
--- a/ci/jjb/jobs/unittests.yaml
+++ b/ci/jjb/jobs/unittests.yaml
@@ -8,7 +8,6 @@
           type: label-expression
           name: node-type
           values:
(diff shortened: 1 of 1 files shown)


ci/jjb/jobs/unittests.yaml
```diff
@@ -8,7 +8,6 @@
           type: label-expression
           name: node-type
           values:
-            - f27-os
```
rochacbruno: shouldn't be replaced with f29-os ?

dkliban: no, pulp 2 does not run on F29

PR #670, opened 2019-05-02
Adding pulpadmin keys to the base_image_ci.yaml playbook

bherrin3:
# Problem

The base `f29-os` cannot be used to replace the `docker-os` template without the appropriate keys loaded into the image.

# Solution

For `f29-os` templates created by `base_image_ci.yaml` to work with pulp-fixtures, the `f29-os` requires the private keys for pulpadmin.

Adding the encrypted pulpadmin private keys to the vault and configuration of template isos.

Moving the `vars` to the correct directory as well in Ansible.

# References
See: https://projects.engineering.redhat.com/browse/SATQE-3467


bherrin3:
# Updates

* Moving the `ci_user` into the role
* Updating permissions in all locations to using octal notation. 

> Typically you want the permissions to be:
> .ssh directory: 700 (drwx------)
> public key (.pub file): 644 (-rw-r--r--)
> private key (id_rsa): 600 (-rw-------)
> lastly your home directory should not be writeable by the group or others (at most 755 (drwxr-xr-x)).

PR #675, opened 2019-06-06
Add vault job for pulp2

rochacbruno:
This job will:
 - Provision a new machine on OpenStack 
 - Install INITIAL_PULP_VERSION (usually one version before the latest). 
 - Install pulp-smash from SMASH_REPO.
- Clone pulp-2-testsfrom TEST_REPO.
- Run py.test sanity check using args provided in PRE_PYTEST_ARGS.
- If it passes then it will upgrade O.S using OS_LATEST_REPO.
- Reboot the instance
- Then it will run the tests again this time using POST_PYTEST_ARGS.

### DIFF
--- a/ci/jjb/jobs/pulp-vault.yaml
+++ b/ci/jjb/jobs/pulp-vault.yaml
@@ -1,0 +1,247 @@
+# Job that help installing Pulp on any machine. This is useful for setting up
+# machines for running Pulp Smash.
+- job:
+    name: 'pulp2-vault'
+    concurrent: true
+    node: 'f29-os'
+    description: |
+        <p>
+          This job will:
+            - Provision an OpenStack instance using image provided in UPSHIFTIMAGE.
+            - Install PULP_VERSION on the running instance.
+            - Install and configure PULP_SMASH using SMASH_REPO.
+            - Clone test cases using TEST_REPO.
+            - If UPGRADE_OS
+                - Run sanity tests defined in PRE_PYTEST_ARGS.
+                - Upgrade O.S using OS_LATEST_REPO.
+            - Run final test suite using POST_PYTEST_ARGS.
+            - Store results.
+        </p>
+
+    parameters:
+        - string:
+            name: UPSHIFTIMAGE
+            default: rhel-7.6-server-x86_64-released
+            description: The name of the image available on Openstack, pulp will be installed on that instance and after sanity check upgrade will be performed.
+        - bool:
+            name: UPGRADE_OS
+            default: true
+            description: Mark if you want to have O.S upgrades, uncheck if you only want to run tests on the O.S without upgrading.
+        - bool:
+            name: DELETE_INSTANCE
+            default: false
+            description: Mark if you want to have the instance deleted after success.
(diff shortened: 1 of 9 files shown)


ci/jjb/scripts/fix_remote_hostname.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+# This information can aid in debugging failed jobs.
+ssh cloud-user@${PULP_HOSTNAME} << EOF
+hostname --all-fqdn
+hostname
+# Setting the hostname of the system to enable connectivity
+hostname="$(hostname --all-fqdn | grep openstacklocal | head -n 1)"
```
bherrin3:
```suggestion
hostname="$(hostname --all-fqdn | grep upshift | head -n 1)"
```

ci/jjb/jobs/pulp-vault.yaml
```diff
@@ -0,0 +1,233 @@
...
+           default: pulp-jenkins
+           description: Can be found on OPenstack user interface.
+        - string:
+           name: UPSHIFTAUTHURL
+           default: https://rhos-d.infra.prod.upshift.rdu2.redhat.com:13000/v3
+           description: Can be found on OPenstack user interface.
+
+        - string:
+            name: PULP_VERSION
+            default: '2.19'
```
bherrin3:
```suggestion
            default: '2.21'
```

:)

ci/jjb/jobs/pulp-vault.yaml
```diff
@@ -0,0 +1,233 @@
...
+        - shell: |
+            # Useful for debugging
+            getent passwd "$(id -u)"
+
+            # Setup ssh config and private key
+            mkdir ~/.ssh/controlmasters
+            
+            cat > ~/.ssh/config <<EOF
+            Host ${PULP_HOSTNAME}
+                User cloud-user
```
bherrin3: The `cloud-user` was just for RHEL. Fedora will use `fedora`. It will be different depending on the base template. :(

rochacbruno: This vault job will always run RHEL

ci/jjb/jobs/macros.yaml
```diff
@@ -56,7 +56,7 @@
             hostname --all-fqdn
             hostname
             # Setting the hostname of the system to enable connectivity
-            hostname="$(hostname --all-fqdn | grep openstacklocal | head -n 1)"
+            hostname="$(hostname --all-fqdn | grep upshift | head -n 1)"
```
bherrin3: @rochacbruno I think if you rebase against `production master`, this has already been merged in.

bherrin3:
Had a quick 1v1 to let @rochacbruno know about DNS problems with the ENV with the current backwards compatible work-around in #682 .

Otherwise, +1 !

ggainey:
This would prob have been useful - but pulp2 is EOL now, so I'm going to close the PR. Thanks for the contribution, sorry it never made it into production :(

PR #677, opened 2019-06-12
Problem: pulp 2 services have been renamed

dkliban:
Solution: update roles with new service names

re: #4554
https://pulp.plan.io/issues/4554


dkliban:
@pulp/qe I did not change the upgrade job because I was not sure how to best check the currently installed version of pulp. Could use some help with writing some ansible that does the same thing as this script for the RPM package: https://github.com/pulp/pulp-packaging/pull/100/files#diff-66fa1680120e21f0a3594e864708059fR499

bherrin3:
@dkliban if the scope and implementation has changed, can we close this out? Thanks!

dkliban:
@bherrin3 , yes, i'll do it now.

PR #678, opened 2019-06-18
Update Pulp 3 ports

kersommoura:
Update Pulp 3 ports to reflect what is used by installer.

See: https://pulp.plan.io/issues/4964
See: https://pulp.plan.io/issues/4965

### DIFF
--- a/ci/jjb/jobs/pulp3-smash-runner.yaml
+++ b/ci/jjb/jobs/pulp3-smash-runner.yaml
@@ -42,7 +42,7 @@
             chmod 600 PRIVATE_KEY ~/.ssh/config
 
             echo "Is pulp working?"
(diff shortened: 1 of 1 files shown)


ci/jjb/jobs/pulp3-smash-runner.yaml
```diff
@@ -65,7 +65,7 @@
                 "hostname": "${PULP_SMASH_SYSTEM_HOSTNAME}",
                 "roles": {
                     "api": {
-                    "port": 80,
+                    "port": 24817,
                     "scheme": "http",
                     "service": "nginx",
```
rochacbruno: This needs change, instead of nginx it is now the `pulp-api.service`

PR #682, opened 2019-07-10
Updates to add hostname to /etc/hosts since DNS is unreliable

bherrin3:
## Problem

PnT DNS for our CI network is not reliable or not returning the fully qualified hostname from DNS.

## Solution
Updating required jobs to pass the IP and HOSTNAME to pulp-smash and pulp-smash to expect those parameters to add to /etc/hosts.

If these parameters are NOT defined, a blank line is added to /etc/hosts.

If DNS IS working, the previous hostname definition should continue to work allowing for previously functioning behavior.

## WIP
Testing on the latest non-functioning DNS as the final test (IP)

### DIFF
--- a/ci/jjb/jobs/pulp-smash-runner.yaml
+++ b/ci/jjb/jobs/pulp-smash-runner.yaml
@@ -8,9 +8,13 @@
     parameters:
         - string:
             name: PULP_SMASH_SYSTEM_HOSTNAME
+            description: A defined name of the smash system to go into /etc/hosts
+        - string:
+            name: PULP_SMASH_SYSTEM_IP
+            description: A defined IPv4 of the smash system to go into /etc/hosts
         - string:
             name: PULP_SMASH_PULP_VERSION
-            decription: The Pulp version Pulp Smash will consider when running tests.
+            description: The Pulp version Pulp Smash will consider when running tests.
         - string:
             name: PULP_SMASH_PULP_USERNAME
             default: admin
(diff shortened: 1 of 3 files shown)


ci/jjb/jobs/pulp-smash-runner.yaml
```diff
@@ -9,6 +9,10 @@
         - string:
             name: PULP_SMASH_SYSTEM_HOSTNAME
         - string:
+            decription: A defined name of the smash system to go into /etc/hosts
+            name: PULP_SMASH_SYSTEM_IP
+        - string:
+            decription: A defined IPv4 of the smash system to go into /etc/hosts
```
kersommoura: There are 2 descriptions for `PULP_SMASH_PULP_VERSION`.

ci/jjb/jobs/pulp-smash-runner.yaml
```diff
@@ -9,6 +9,10 @@
         - string:
             name: PULP_SMASH_SYSTEM_HOSTNAME
         - string:
+            decription: A defined name of the smash system to go into /etc/hosts
```
kersommoura:
```suggestion
            description: A defined name of the smash system to go into /etc/hosts
```

Should be IP?

bherrin3: A lot of weird cut and paste happened. Resolving.

PR #684, opened 2019-09-06
Update dev ownership

kersommoura:
Update dev ownership, and add devs to the pulp-dev jobs.


kersommoura:
@dkliban has agreed with those changes. Merging this PR.

PR #687, opened 2019-09-09
Fix ownership for Pulp 2 dev job

kersommoura:
Fix ownership for Pulp 2 dev job. Creates `pulp2-ownership` property to
be shared among QE and DEV team.


kersommoura:
@bherrin3, any comments?

PR #706, opened 2020-03-24
Exclude bot PRs from pr processor

daviddavis:
[noissue]

### DIFF
--- a/.github/pr_processor.py
+++ b/.github/pr_processor.py
@@ -39,15 +39,20 @@ regex = r"(?:{keywords})[\s:]+#(\d+)".format(keywords=("|").join(KEYWORDS))
 pattern = re.compile(regex)
 
 for repo in REPOS:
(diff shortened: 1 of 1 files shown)


.github/pr_processor.py
```diff
@@ -39,17 +39,22 @@ def user_comment(issue, user=GITHUB_USER):
...
     for issue in issues:
         print(f"Processing issue {issue.number}")
+
         # check if we've seen this PR already
         if user_comment(issue):
             print(f"Issue {issue.number} already processed. Skipping.")
             continue
 
+        # check if our bot opened this issue
+        if issue.user.login == GITHUB_USER:
```
fabricio-aguiar: Is it due to cherrypick right?

daviddavis: Yea, that's the only case right now.

PR #710, opened 2020-05-14
Adding more repos to PR processor

fao89:
- pulp_ansible
- pulp_deb
- pulp-certguard
- pulp_installer
- pulp_rpm_prerequisites

pulp_installer and pulp_rpm_prerequisites not adding cherrypick label

### DIFF
--- a/.github/pr_processor.py
+++ b/.github/pr_processor.py
@@ -7,7 +7,17 @@ from github.GithubException import UnknownObjectException
 from redminelib import Redmine
 
 ORG = "pulp"
-REPOS = ["pulpcore", "pulp_file", "pulp_rpm", "pulp_container"]
+REPOS = [
+    "pulpcore",
+    "pulp_file",
+    "pulp_rpm",
+    "pulp_container",
+    "pulp_ansible",
+    "pulp_deb",
+    "pulp-certguard",
+    "pulp_installer",
+    "pulp_rpm_prerequisites",
+]
 SINCE = datetime.utcnow() - timedelta(hours=1)
 KEYWORDS = ["fixes", "closes", "re", "ref"]
 PR_STATUS = 3  # POST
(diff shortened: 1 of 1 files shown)


.github/pr_processor.py
```diff
@@ -7,7 +7,16 @@
...
-REPOS = ["pulpcore", "pulp_file", "pulp_rpm", "pulp_container"]
+TOOLS = ["pulp_installer", "pulp_rpm_prerequisites"]
+REPOS = [
+    "pulpcore",
+    "pulp_file",
+    "pulp_rpm",
+    "pulp_container",
+    "pulp_ansible",
+    "pulp_deb",
+    "pulp-certguard",
```
daviddavis, line 18: Did you check with pulp_deb and pulp-certguard teams?

.github/pr_processor.py
```diff
@@ -79,7 +88,7 @@ def user_comment(issue, user=GITHUB_USER):
             for issue_num in r_issues:
                 r_issue = redmine.issue.get(issue_num)
 
-                if r_issue.tracker.name == "Issue":
+                if r_issue.tracker.name == "Issue" and repo not in TOOLS:
```
daviddavis, line 91: If these repos don't have cherry pick tags, then it should just ignore them.

daviddavis: See https://github.com/pulp/pulp-ci/blob/master/.github/pr_processor.py#L98-L102

.github/pr_processor.py
fao89: No, I didn't

.github/pr_processor.py
fao89: cool, I didn't notice it before, thanks!

.github/pr_processor.py
fao89: @mdellweg @quba42 I'm adding pulp_deb here, it will enable @pulpbot to update redmine to post automatically, and add cherry-pick label if the PR refers to an issue.

If you don't want it, just ping me and I will remove it

quba42: Personally I was going to wait a bit longer before bringing this up.
That being said, I am fine with adding it now and doing most work with a pulp.plan.io issue.

The final call should be made by @mdellweg.

PR #712, opened 2020-05-27
Fixing xmlrpc import

fao89:


daviddavis:
I ran 2to3 and it looks like there are a couple other small items that it tried to fix:

```
--- ci/redmine_bugzilla.py	(original)
+++ ci/redmine_bugzilla.py	(refactored)
@@ -115,7 +115,7 @@
                 bug = BZ.getbug(bug_id)
             except Fault as e:
                 if e.faultCode == 102:
-                    print("Bugzilla %s could not be accessed." % bug_id)
+                    print(("Bugzilla %s could not be accessed." % bug_id))
                     continue
                 else:
                     raise
@@ -131,7 +131,7 @@
                         ext_params["ext_status"] = issue.status.name
                     if external_bug["ext_priority"] != issue.priority.name:
                         ext_params["ext_priority"] = issue.priority.name
-                    if len(ext_params.keys()) > 0:
+                    if len(list(ext_params.keys())) > 0:
                         ext_bug_record += (
                             "Bugzilla bug %s updated from upstream bug %s "
                             "with %s\n" % (bug.id, issue.id, ext_params)
RefactoringTool: Files that need to be modified:
RefactoringTool: ci/redmine_bugzilla.py
```

PR #713, opened 2020-05-27
Changing downstream contact emails

daviddavis:


ci/redmine_bugzilla.py
```diff
@@ -35,7 +35,7 @@
 BUGZILLA_URL = "https://bugzilla.redhat.com"
 REDMINE_URL = "https://pulp.plan.io"
 
-DOWNSTREAM_CONTACTS = ["dkliban@redhat.com", "ttereshc@redhat.com"]
+DOWNSTREAM_CONTACTS = ["daviddavis@redhat.com", "bmbouter@redhat.com"]
```
daviddavis, line 38: @dkliban @goosemania please let me know if you want to remain contacts.

@bmbouter please let me know if you are not ok with being emailed.

fao89: it is hardcoded here also:
https://github.com/pulp/pulp-ci/blob/master/.github/send_email.py#L64-L70

daviddavis: Should the send_email.py addresses be different than the ones here (ie do we imagine that the send_email addresses will be used for other jobs besides this script)?

fao89: for now, it is only used by redmine CI job, it was introduced just for it.
It was thought for CI to use to deliver emails, maybe in the future, we will use it for other jobs

PR #714, opened 2020-05-27
Adding some temporary output to debug redmine/bz automation

daviddavis:


ci/redmine_bugzilla.py
```diff
@@ -104,6 +104,7 @@ def main():
     failed_qa_bugzillas = []
 
     for issue in redmine_issues:
+        print(f"Processing {issue.id}.")
```
fao89, line 107: Cool!
Maybe printing a `len(redmine_issues)` could be useful

PR #729, opened 2020-08-17
Run the spam cleanup script as a cron job

lubosmj:
In this commit, there is decreased the threshold that specifies how much we should look into history along with the time required for the response.


lubosmj:
Do not merge this pull request until further notice.

PR #741, opened 2020-10-30
Prevent smtplib errors caused by gmail's security policies

lubosmj:
In this commit, emails are sent with one second pause between them. This change was made due to the following exception: "smtplib.SMTPDataError: 421, 4.3.0 Temporary System Problem.  Try again later (10)."

Another change had to be made in the process of updating the sheet. As of now, the sheet is updated only after when the procedure which sends notifications to unverified users succeeds. This could cause a state where a user was not notified and his/her content was deleted silently.

One problem still remains. From time to time, the script fails with the error: "smtplib.SMTPAuthenticationError: (534, b'5.7.14 <https://accounts.google.com/signin/continue?XXXXXXXXXX Please log in via your web browser and then try again.\n5.7.14  Learn more at\n5.7.14  https://support.google.com/mail/answer/78754 - gsmtp')". To cope with that, there was added a second trial for logging in. However, the logging process may fail again and this cannot be resolved from the script's perspective.

### DIFF
--- a/.github/spam_cleaner.py
+++ b/.github/spam_cleaner.py
@@ -5,8 +5,10 @@ import logging
 import logging.config
 import os
 import pickle
+import random
 import re
 import smtplib
+import time
 import tldextract
 
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


bmbouter (approved):
This looks fine to me.

PR #742, opened 2020-10-30
Start tagging the PR processor comment

daviddavis:
This will be useful when the PR processor has to revisit PRs and find its comment. See https://pulp.plan.io/issues/6720 as an example.


daviddavis:
We're moving to Github Issues which will not use our pr processor. Closing.

PR #749, opened 2021-03-05
Add migration plugin badges to the README

goosemania:
[noissue]

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,3 +78,8 @@ This repository is the home of Pulp's CI files.
 * [![Pulp Nightly CI/CD](https://github.com/pulp/pulp_python/actions/workflows/nightly.yml/badge.svg)](https://github.com/pulp/pulp_python/actions/workflows/nightly.yml)
 * [![Pulp Release CI/CD](https://github.com/pulp/pulp_python/actions/workflows/release.yml/badge.svg)](https://github.com/pulp/pulp_python/actions/workflows/release.yml)
 
+### Pulp 2to3 Migration
+
+* [![Pulp CI](https://github.com/pulp/pulp-2to3-migration/actions/workflows/ci.yml/badge.svg?branch=master)](https://github.com/pulp/pulp-2to3-migration/actions/workflows/ci.yml)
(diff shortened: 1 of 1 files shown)


ipanova:
looking at other plugins' badges i see also the FIPS status, does it makes sense to add it to the migration plugin as well?

goosemania:
> looking at other plugins' badges i see also the FIPS status, does it makes sense to add it to the migration plugin as well?

migration plugin doesn't have fips job at the moment

ipanova:
> > looking at other plugins' badges i see also the FIPS status, does it makes sense to add it to the migration plugin as well?
> 
> migration plugin doesn't have fips job at the moment

I'm aware of that. It will eventually.

goosemania:
> > > looking at other plugins' badges i see also the FIPS status, does it makes sense to add it to the migration plugin as well?
> > 
> > 
> > migration plugin doesn't have fips job at the moment
> 
> I'm aware of that. It will eventually.

I do not know which path to the workflow to use if it doesn't exist yet. And I'm not sure that the standard one used by other plugins will be used in the migration, because it will likely have a different box and potentially a different workflow.
What do you want me to do with this PR?

ipanova:
> > > > looking at other plugins' badges i see also the FIPS status, does it makes sense to add it to the migration plugin as well?
> > > 
> > > 
> > > migration plugin doesn't have fips job at the moment
> > 
> > 
> > I'm aware of that. It will eventually.
> 
> I do not know which path to the workflow to use if it doesn't exist yet. And I'm not sure that the standard one used by other plugins will be used in the migration, because it will likely have a different box and potentially a different workflow.
> What do you want me to do with this PR?

my suggestion was to add this line 
`* [Pulp Migration FIPS CI](https://github.com/pulp/pulp-2to3-migration/actions/workflows/fips.yml)`
but i have not tested when a badge is being provided a non existent path, maybe it's a problem. hence it was purely a suggestion with intention to spare a follow up pr.

ipanova (approved):

fao89:
> looking at other plugins' badges i see also the FIPS status, does it makes sense to add it to the migration plugin as well?

actually, our FIPS CI times out, so we will probably remove it soon

PR #756, opened 2021-05-25
Update supported releases

dralley:
[noissue]


ci/config/releases/supported-releases.json
```diff
@@ -1,5 +1,4 @@
 [ 
-    "2.20",
```
dralley, line 2: I've seen conflicting information about this, but it looks like 6.7 is on 2.21?  And since 6.6 is EOL as of a few weeks ago, we no longer need 2.20 if that is true.

mdellweg (approved):

PR #833, opened 2023-05-23
Adds the `oci_env` CI status badges.

decko:
[noissue]


decko:
Makes no sense since we don't have a `release` workflow yet. Closing it.

PR #867, opened 2024-09-11
Update supported versions

pulpbot:
[noissue]


ggainey:
Nothing uses this any more

